PR #9, opened 2014-07-11
Allows a term to appear more than once in one term.

bobot:
One sometimes need the value of one option when e sanitize another option. For clarity it is interesting to separate the sanitization of each options when each one is defined. However the current implementation of cmdliner doesn't support to use an option more than once in the same term. This merge request allows to use more than once the same term in one term:
        - the computed value is memoized;
        - it appears only once in the documentation;
        - add an example that use this feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 \.\#*
 \#*#
 CLOCK.org
+*.native
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -124,6 +124,40 @@ end = struct
   let of_list l = List.fold_left (fun t (s, v) -> add t s v) empty l
 end
 
+module Hetero_Hashtbl : sig
+  type 'a tag
+  type t
+
+  val create: int -> t
+  val create_tag: unit -> 'a tag
+  val int_of_tag: 'a tag -> int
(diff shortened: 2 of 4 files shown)


Drup:
I have an interest in this feature as well. I don't have much comment on the implement than yours, however, on the semantics aspect, I don't really see issues. Either you specify that side effects are evaluated each time or you specify they are evaluated only once. The semantic is rather clear anyway.

Personally, what *really* matter (and that I would consider a bug in the current implementation) is that: 
- The argv parsing should be robust to double declaration of options.
- The documentation should not show options twice.

dbuenzli:
> Do you think you will take another look at it?

I may have a look at this again when I work on cmdliner in the future (i.e. not now). 

> (Will you accept the dependency on your own library ;)

I don't think so, a lot of people are using `cmdliner` and the deal so far was that it had no dependency, Also `hmap` is a little bit cutting edge (4.02.0), `cmdliner` until now is 3.12.1, though the latter will have to be lifted at some point.

dbuenzli:
So I think I'll aim for @drup's [comment](https://github.com/dbuenzli/cmdliner/issues/77) with side-effects evaluated each time but decent documentation. This is tracked in https://github.com/dbuenzli/cmdliner/issues/77

PR #17, opened 2014-11-30
Add `vopt` optional argument type

nojb:
This is an analog of `vflag` but for optional arguments that carry a
value.  ~~As far as I can see it is not possible to obtain this
functionality without modifying the library.~~  I need this because I am
generating a Cmdliner term dynamically.

The name `vopt` clashes with the optional argument `?vopt`, so maybe a
different one is needed.

**UPDATE** This is very easily done using the `app` combinator on the level of terms (instead of on the level of arguments).

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -945,7 +945,7 @@ module Arg = struct
     let al = List.map (fun (v, (_, print), a) ->
         if is_pos a then invalid_arg err_not_opt
         else { a with absent = Val (lazy (str_of_pp print v)); o_kind = Opt }) l in
-    let convert_one cl (v, (parse, print), _) a =
+    let convert_one cl (v, (parse, _), _) a =
       match Cmdline.opt_arg cl a with
       | [] -> v
       | [_, f, Some v] -> parse_opt_value parse f v
--- a/test/dynargs.ml
+++ b/test/dynargs.ml
@@ -1,0 +1,25 @@
+open Cmdliner
+
+let getenv n =
+  try Sys.getenv n with Not_found -> ""
+
+let main () =
+  let args =
+    let parse = fst Arg.(list ~sep:',' (pair ~sep:':' string int)) in
+    match parse (getenv "ARGS") with
+    | `Ok args -> args
+    | `Error _ ->
+        prerr_endline
+          "Error: $ARGS could not be parsed, should be \
+           of the form <name1>:<int1>,<name2>:<int2>,...";
+        exit 1
+  in
+  let args = List.map (fun (n, k) -> (k, Arg.int, Arg.info [ n ])) args in
+  if List.length args = 0 then
+    prerr_endline "No arguments specified.  Try setting $ARGS: `ARGS=foo:0,bar:101 ./dynargs.native`"
+  else
+  let cmd = Term.(pure (fun _ -> `Help (`Pager, None)) $ Arg.(value & vopt args)) in
+  ignore (Term.(eval (ret cmd, info "dynargs")))
+
+let _ =
+  main ()
--- a/test/tests.itarget
+++ b/test/tests.itarget
@@ -2,5 +2,6 @@ revolt.native
 chorus.native
 cp_ex.native
 tail_ex.native
-darcs_ex.native	
+darcs_ex.native
 rm_ex.native
+dynargs.native


tekknolagi:
@nojb sorry to bring up a long-dead thread, but what do you mean by "this is easily done using the app combinator..." -- do you by chance have an example?

nojb:
Hi @tekknolagi, unfortunately I can't decipher what I was trying to do at the time from (my own) description.  Maybe if you explain what you are trying to do, we could discuss that.  Thanks!

tekknolagi:
@nojb Sounds great! I am trying to have this kind of behavior: the user can pick one and only one of:

`--output [filename]`

OR

`--in-place`

but if neither is specified, it defaults to something else.

I have set up the datatype as

```ocaml
type output_type = File of string | InPlace | Stdout
```

nojb:
See below for a possible approach:

```ocaml
open Cmdliner

type output_type =
  | File of string
  | InPlace
  | Stdout

let output_t =
  Arg.(value (opt ~vopt:(Some "") (some string) None (info ["output"])))

let in_place_t =
  Arg.(value (flag (info ["in-place"])))

let t def =
  let merge output in_place =
    match output, in_place with
    | Some _, true -> failwith "can only use one of the options --output, --in-place"
    | Some "", false -> Stdout
    | Some s, false -> File s
    | None, true -> InPlace
    | None, false -> def
  in
  Term.(const merge $ output_t $ in_place_t), Term.info "foo"

open Printf

let print = function
  | Stdout -> print_endline "Stdout"
  | File s -> Printf.ksprintf print_endline "File %S" s
  | InPlace -> print_endline "InPlace"

let () =
  match Term.eval (t Stdout) with
  | `Error _ -> exit 1
  | `Ok v -> print v
  | _ -> exit 0
```

tekknolagi:
@nojb That looks like roughly what I have now -- but I am curious how well that scales to, say, 10 different flags aside from `--output` and `--in-place`

nojb:
Well, the only ugly part is the `merge` function which is only there to require that exactly one of `--output` or `--in-place` is passed.  The code could be made shorter if you simply decree that the latest option passed takes precedence.

As for having 10 different flags, well then you will need to define 10 different `Term.t`'s, I don't think there is much to say about that, and don't see any problem with the way that "scales" ...

Gbury:
Well, in the case of 10 different flags, the `merge` function would become quite monstrous (pattern-matching on a 10-uplet) don't you think ?

nojb:
The point is that typically a `merge` function is not needed at all.  I only included it in the example to show how one could require that exactly one of two different options is passed, but this is not a usual requirement and is not normally needed (rather, last-takes-precedence or other such criteria is typically used), so no such function would typically be needed .

tekknolagi:
If I wanted last-takes-precedence, how would I go about that?

nojb:
You can use `Cmdliner.Arg.last` for this, but it may not fit exactly with what you want to do (mixing flags and options).  I guess this library does not offer a systematic way to handle priorities between different flags/options... but it is not clear that it is a good idea to design command line parameters so that they depend on such a prioritization mechanism.

For example, in your case you could simply say that if no `--output [file]` option is passed then you default to the `--in-place` behaviour, etc, so that both behaviours are dealt with a single option, instead of two.  In short, one should design command line parameters so that they are *orthogonal*.

tekknolagi:
Well they *are* orthogonal, no? If none specified, output to stdout. If `--in-place` specified, write to the file given as the input parameter. If `--output` specified, write to the file given. They can't coexist. As for other conflicts, like reading from stdin and `--in-place`, I handle that elsewhere.

nojb:
IMO, orthogonal does not mean "can't coexist", but rather the opposite: that they can be composed in any fashion, without any restriction, each concept completely independent from the others.  In your situation, `--in-place` and `--output` are not orthogonal in this way.

tekknolagi:
Oh I misunderstood you -- my bad. I think you make a good point. Thank you.

dbuenzli:
@nojb's initial proposal is the right way to solve these problems, except he should not use `failwith` to report errors but a value of type [`Term.ret`](http://erratique.ch/software/cmdliner/doc/Cmdliner.Term#TYPEret) value to be erased with the [`Term.ret`](http://erratique.ch/software/cmdliner/doc/Cmdliner.Term#VALret) combinator.

PR #21, opened 2015-01-30
build: don't require ocamlfind.

AltGr:
Ref: https://github.com/OCamlPro/opam-depext/issues/2


dbuenzli:
Thanks the patch is in as f699105f79f536f70b9.

PR #26, opened 2015-03-25
Add ?fmt to Term.eval and Term.eval_choice

dra27:
I don't "like" the default option of `pager` for `--help` (especially when the help screen is considerably shorter than my terminal, switching to less is tedious). This patch adds `?fmt` to `Term.eval` and `Term.eval_choice` which gets passed on to `add_std_opts`.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -1216,7 +1216,7 @@ module Term = struct
   let remove_exec argv =
     try List.tl (Array.to_list argv) with Failure _ -> invalid_arg err_argv
 
-  let add_std_opts ei =
+  let add_std_opts ei def_fmt =
     let docs = (fst ei.term).sdocs in
     let args, v_lookup =
       if (fst ei.main).version = None then [], None else
@@ -1230,15 +1230,15 @@ module Term = struct
         let fmt = Arg.enum ["pager",`Pager; "groff",`Groff; "plain",`Plain] in
         let doc = "Show this help in format $(docv) (pager, plain or groff)."in
         let a = Arg.info ["help"] ~docv:"FMT" ~docs ~doc in
-        Arg.opt ~vopt:(Some `Pager) (Arg.some fmt) None a
+        Arg.opt ~vopt:(Some def_fmt) (Arg.some fmt) None a
       in
       List.rev_append a args, lookup
     in
     h_lookup, v_lookup,
     { ei with term = (fst ei.term), List.rev_append args (snd ei.term) }
 
-  let eval_term help err ei f args =
-    let help_arg, vers_arg, ei = add_std_opts ei in
+  let eval_term help fmt err ei f args =
+    let help_arg, vers_arg, ei = add_std_opts ei fmt in
     try
       let cl = Cmdline.create (snd ei.term) args in
       match help_arg ei cl, vers_arg with
@@ -1260,19 +1260,20 @@ module Term = struct
             {ei with term = cmd }
         | None -> { ei with term = ei.main }
         in
-        let _, _, ei = add_std_opts ei in
+        let _, _, ei = add_std_opts ei fmt in
         Help.print fmt help ei; `Help
 
   let eval ?(help = Format.std_formatter) ?(err = Format.err_formatter)
-      ?(catch = true) ?(argv = Sys.argv) ((al, f), ti)  =
+      ?(catch = true) ?(argv = Sys.argv) ?(fmt = `Pager) ((al, f), ti)  =
     let term = ti, al in
     let ei = { term = term; main = term; choices = [] } in
-    try eval_term help err ei f (remove_exec argv) with
+    try eval_term help fmt err ei f (remove_exec argv) with
     | e when catch ->
         Err.pr_backtrace err ei e (Printexc.get_backtrace ()); `Error `Exn
 
   let eval_choice ?(help = Format.std_formatter) ?(err = Format.err_formatter)
-      ?(catch = true) ?(argv = Sys.argv) (((al, f) as t), ti) choices =
+      ?(catch = true) ?(argv = Sys.argv) ?(fmt = `Pager) (((al, f) as t), ti)
+      choices =
     let ei_choices = List.rev_map (fun ((al, _), ti) -> ti, al) choices in
     let main = (ti, al) in
     let ei = { term = main; main = main; choices = ei_choices } in
@@ -1281,7 +1282,7 @@ module Term = struct
       let find_chosen (_, ti) = ti = chosen in
       let (al, f), _ = List.find find_chosen ((t, ti) :: choices) in
       let ei = { ei with term = (chosen, al) } in
-      eval_term help err ei f args
+      eval_term help fmt err ei f args
     with
     | Cmdline.Error e ->                    (* may be raised by choose_term. *)
         Err.pr_usage err ei e; `Error `Parse
@@ -1293,7 +1294,7 @@ module Term = struct
     let version = if version_opt then Some "dummy" else None in
     let term = info ?version "dummy", al in
     let ei = { term = term; main = term; choices = [] } in
-    let help_arg, vers_arg, ei = add_std_opts ei in
+    let help_arg, vers_arg, ei = add_std_opts ei `Pager in
     try
       let cl = Cmdline.create ~peek_opts:true (snd ei.term) args in
       match help_arg ei cl, vers_arg with
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -172,7 +172,8 @@ module Term : sig
 
   val eval : ?help:Format.formatter ->
     ?err:Format.formatter -> ?catch:bool ->
-    ?argv:string array -> ('a t * info) -> 'a result
+    ?argv:string array -> ?fmt:[`Pager | `Plain | `Groff ] -> ('a t * info) ->
+    'a result
   (** [eval help err catch argv (t,i)]  is the evaluation result
       of [t] with command line arguments [argv] (defaults to {!Sys.argv}).
 
@@ -182,12 +183,14 @@ module Term : sig
 
       [help] is the formatter used to print help or version messages
       (defaults to {!Format.std_formatter}). [err] is the formatter
-      used to print error messages (defaults to {!Format.err_formatter}). *)
+      used to print error messages (defaults to {!Format.err_formatter}).
+
+      [fmt] specified the default argument for --help ([`Pager]). *)
 
   val eval_choice : ?help:Format.formatter ->
     ?err:Format.formatter -> ?catch:bool ->
-    ?argv:string array -> 'a t * info -> ('a t * info) list ->
-    'a result
+    ?argv:string array -> ?fmt:[`Pager | `Plain | `Groff ] -> 'a t * info ->
+    ('a t * info) list -> 'a result
   (** [eval_choice help err catch argv default (t,i) choices] is like {!eval}
       except that if the first argument on the command line is not an option
       name it will look in [choices] for a term whose information has this


dbuenzli:
I'm not very fond of over-configurability. `cmdliner` tries to solve the average use case well, in the interest of the end user. 

Most command line tools will have more than a screenful of documentation (or will quickly evolve to such a state or if they don't, maybe they are underdocumented...) in which case you most often want paging for search. My experience with tools which default to non-paging behaviour is that you end up invoking the tool twice, once with `--help` and immediately after with `--help | less` which is needlessly annoying.

So in my opinion not letting users of the library tweak this will result in a more uniform and "correct" user experience for the end-user.

dsheets:
Sometimes I've wanted a `-h` option which gives a brief summary to `stdout` of the subcommands or major options. I've never wanted it enough to attempt to support it, though. If I got this functionality for free, somehow, I wouldn't complain, though.

dra27:
I think it's more personal opinion than good practice. As a *user* - I find the default piping to another command a bit un-UNIXy (and I'm running on Windows :wink:). Perhaps it's because my expectation of --help is different from how more modern systems seem to be doing it - I expect --help to be no more than a screen or two's information (so well within the scroll-back of my terminal/prompt - I certainly don't end up double-running commands) and to run something completely different (typically man foo or info foo) if the information runs to pages. I find git's help output tedious for this reason, but perhaps I'm being a dinosaur! I do however find losing the entire content of my terminal (and the flicker) to be amateur for displaying a man page which only occupies 25% of its viewable height. This particular tool, FWIW, will get more complicated, but its command line arguments won't change...
I guess that attempting to have cmdliner detect if the terminal is large enough to display the output without using a pager is too big a can of worms.
Would it be more acceptable as part of a more general mechanism to override the --help (e.g. also allowing the sentence for the help text to be changed and, even, to omit certain options - groff, for example, is less useful on a Windows system)?

dbuenzli:
> Sometimes I've wanted a -h option which gives a brief summary to stdout of the subcommands or major options.

I can understand that, but then you need to be able to specify in the API what goes into that summary and I'm not necessarily fond of multiplying the documentation entry points/forms. In my opinion these summaries don't perform especially well when you are after something –– especially if the thing you look for is not in the summary. I rather have exhaustiveness directly and search at my finger tip (and hence paging).

As for `stdout`, always the same problem for me that's a typical (annoying) interaction I will have with git (on a 80x24 terminal):
```
git --help            # realise output is truncated
git --help | less   # realise my command is not in the summary
git help -a          # realise output is truncated
git help -a | less 
```

dbuenzli:
> Would it be more acceptable as part of a more general mechanism to override the --help (e.g. also allowing the sentence for the help text to be changed and, even, to omit certain options - groff, for example, is less useful on a Windows system)?

I see little value in allowing that but adding more cruft to the API. I don't see having a standard `--help` with standard options (and standard description) which work across platforms as being a curse, I prefer if the tools work uniformly across platforms. 

It seems not all users expect/like to have manuals on `--help` but it's a pattern I have seen increasingly being used by command line tools and I think that from a documentation access perspective it's a better pattern (if you have search) than short summaries on `--help` + a man page that may be missing or not in your `MANPATH`. So that's the pattern cmdliner chooses to support.

dra27:
Internationalisation? The `--help` option is already inconsistent across platforms, so I'm not quite sure where you're going with that. I guess my "problem" here is that your coding library is trying to influence how **my** command line works and on a matter which just personal opinion, but that's of course completely your call! Maintaining a fork is not exactly hard.

dbuenzli:
Le jeudi, 26 mars 2015 à 10:16, David Allsopp a écrit :
> Internationalisation?

I think that internationalising command line tools it one of the stupidest idea out there.

> my "problem" here is that your coding library is trying to influence how my command line works and on a matter which just personal opinion,  
It's not trying to influence how your command line works, it *defines* how your command line works so that you don't even have to think about this. And it does this in the way that I expect to be the best for the end-user. Design is about making choices, not about trying to please everyone.

Best,

Daniel

rleonid:
@dbuenzli  If I understand your https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner.mli#L103

dbuenzli:
@rleonid ?

rleonid:
Apologies, premature `ctrl+enter`

dbuenzli:
> Personally, I prefer this pattern because the terminal that I use has built in search.

Search is not the only issue. Avoiding window scrolling is another. Usually I'm interested in starting reading from the start of the page, not getting the end of the page and having to scroll back at the beginning.

> git being the first that comes to mind.

...and I highlighted the kind of terrible user experience you get with git's help above.

> Ultimately, I think the decision should come down to the tool implementer

Well I don't think so, I'm tired of each tool at hand operating in a different manner, so I think a cli parsing library is the right place to standardize this behaviour, regardless of the what the tool implementer thinks about the matter. The behaviour you is bad for end-users since cmdliner doesn't output summaries (and has not plans to do so, see again discussion above) but a plain and long man page, forcing in most cases your users to scroll back to a not so well defined point, or to invoke the same command a second time by piping the result into a pager.

The only compromise I'd be willing to do is to default to a more sensible value according to end-user expectations on a given *platform* if there are standardized human interface guidelines w.r.t. behaviour.
 
> how would you feel about a PR that extended the variant https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner.mli#L103 with something like `Stdout that writes the man page to stdout?

That's what `Plain or `Groff do. 

Apart from this sensible request https://github.com/dbuenzli/cmdliner/issues/43 which I'm going to fix in a future release, I have so far received little complaints about this behaviour by end-users, so I'm not planning to change this.

rleonid:
> That's what Plain or Groff do. 
I meant without the pager.

> I have so far received little complaints about this behaviour by end-users, so I'm not planning to change this.

It seems like you've received at least 2 PR's to change this. There is a selection bias to the feedback that you're receiving.

Would you welcome a PR to #43 ? End users can then create processes that fork themselves and modify their child's environment.

dbuenzli:
> It seems like you've received at least 2 PR's to change this.

Well the first one was for aesthetics reasons and the second one for doubtful usability reasons which I counter point by point in this discussion. 

So far nobody gave me good *usability* reasons to prevent the current behaviour --- it was only individual pet peeves --- and end-users do not complain, so there's little incentive to change anything.

> I meant without the pager.

There will be no pager if you use Plain or Groff, it will output to stdout, it's what you get with `--help=plain` or `--help=groff`. But you can't override what --help will do.

> Would you welcome a PR to #43 ?

Why not, but IIRC last time I looked it was not so easy to change. You can't lookup the env at the point I mentioned in the issue and use as the default as it will change the default value and the way it is printed in the manual and later you don't know if the value you get was specified on the cli or not (related to #52). This is due to the whole internal disgusting crap that cmdliner's implementation is which I'd like to redo at some point. So if you manage to find a way and do a PR don't be surprised if it gets overwritten before it's actually released.

> End users can then create processes that fork themselves and modify their child's environment.

You mean users of cmdliner not end-users right ? Don't do this, you'll loose windows compat.

Ulrar:
This whole discussion is honestly baffling. Having to work around a library and use dumb things like `binary --help | cat` to get the behavior you want is insane, just because the developer doesn't like giving options ? The subject isn't even changing the default here, it's literally just giving the choice.

Seems like I'm far from the only one to be annoyed by this, and I'm in the vast, vast minority of people who actually know what cmdliner is. Giving the option literally doesn't cost you anything, and people will stop opening issues about you trying to force your own preference down their throats, sounds good to me.

dbuenzli:
> The subject isn't even changing the default here, it's literally just giving the choice.

I'm sorry you don't understand that choice is not always good. 

It's a little bit sad that you did not bring anything of value to this discussion. If you pay attention to the points made above you will understand why cmdliner operate this way and that it's better if *you* don't try to force your own preferences down the throats of cli users.

dbuenzli:
Just to make the point more obvious. In the current state of the `opam-repository` we have: 
```
> opam list --depends-on cmdliner | wc -l 
     202
```
I hope you can understand that for user experience consistency it is undesirable for end-users if each of these package authors makes a choice and forces down its own preference down the throats of its users.

dra27:
@Ulrar - regardless of one’s opinion, the discussion is indeed long-since done. It’s an open source world - the patch should be easy to rebase and and so you can vendor cmdliner if the feature is sufficiently important to you. The Dune overlays, if you use Dune, would make that even less painful.

Ulrar:
You are assuming a lot of things, firstly not everyone has end-users (I'm my own end user, I'm annoyed by this : an end user just complained).

I did read the thread, and others, and I remain as unconvinced by your reasoning as you are by ours (emphasis on the plural). Your opinion on this isn't more relevant than anyone else's, you get to dictate because it's your lib, and granted you do whatever you want with your own library but don't try to sell it as anything else than imposing your own preference on others for no reason, because that's what it is.

Sure, there's a lot of users of cmdliner in the ocaml world, great. But this is very inconsistent with most other binaries, as a sysadmin I spend my day using a terminal and the little app I made with cmdliner yesterday is literally the only one I have a problem with. Which makes your point nil imho, cmdliner is actually forcing an inconsistent default on users.

Anyway, this has clearly been going on for years, no point in arguing over something you won't fix, I'll parse my arguments some other way.

avsm:
@Ulrar wrote:
> Seems like I'm far from the only one to be annoyed by this, and I'm in the vast, vast minority of people who actually know what cmdliner is. Giving the option literally doesn't cost you anything, and people will stop opening issues about you trying to force your own preference down their throats, sounds good to me.

Putting aside the disrespectful language and tone you're adopting, I agree with @dbuenzli and think you're wrong.  The few users that really want a pager-free experience already have the `--help=plain` option.

A more "UNIX-y" way of solving the problem for a better user experience would be to add Dune rules to generate proper manual pages from the command during build time, and install those so that they are available for search via `man -k` as well.

Ulrar:
> Putting aside the disrespectful language and tone you're adopting, I agree with @dbuenzli and think you're wrong. The few users that really want a pager-free experience already have the `--help=plain` option.

Sorry about this then, didn't mean to be disrespectful

dmtrKovalenko:
@dbuenzli 
> Most command line tools will have more than a screenful of documentation (or will quickly evolve to such a state or if they don't, maybe they are underdocumented...) in which case you most often want paging for search. My experience with tools which default to non-paging behaviour is that you end up invoking the tool twice, once with `--help` and immediately after with `--help | less` which is needlessly annoying.
> 
> So in my opinion not letting users of the library tweak this will result in a more uniform and "correct" user experience for the end-user.

This is completely incorrect. Most people are confused by pager output and trying to close and reopen the help message without a pager. Moreover, it doesn't support colors and format in the output, which is completely important. The default option definitely needs to be `plain` and if the user wants - he can type `--help=pager`

dra27:
As it happens I was thinking of this last week owing to an unrelated piece of work. Certainly my original issue - and possibly yours - would be fixed if by default `LESS` were set to `FRX` as Git does (see docs for [core.pager](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corepager).

I don’t know if replicating its pager behaviour is a good or bad thing, just that it’s a “lot of users regularly using a pager” thing 🙂

dmtrKovalenko:
Actually my main problem is the term colons in output. Colours in help messages, especially in the long messages are soo important. And it is just not possible to include them by default.

dbuenzli:
I have never seen someone actively trying to "close" `git log` to get a non paged output… 

In any case the help message in cmdliner is a man page and the only distinctions we have there is normal, bold and italics text. These distinctions render fine as such here, isn't that the case for you ? 

The `FRX` dance can certainly [be done](https://github.com/b0-system/b0/blob/96224e38bf8a70d4ea632114aa09fbb965239d3d/src-b00/kit/b00_pager.ml#L53-L57) but it would be difficult to make is solely based on `Sys`. But given the man page output I'm not sure what this would bring to the party.

dmtrKovalenko:
`git log` is not the `--help` message. The pager name itself means that you will have several pages. 

But help message is not a giant infinity list. It's a small guide of how to use a library. It should read fast and be clear (that's why colors are important)

dra27:
Actually, it’s `git diff` I’m thinking of - if the diff is a smaller than the terminal, then the pager exits and you don’t have to quit it yourself (which is where this all started!)

dra27:
@dmtrKovalenko - a rebased version of this PR lives at https://github.com/dra27/cmdliner/tree/custom-help. The rehashing of the upstreaming discussion doesn’t seem worthwhile unless there’s something new to add - @dbuenzli answered the choice argument in https://github.com/dbuenzli/cmdliner/pull/26#issuecomment-215135858

dbuenzli:
@dmtrKovalenko what you want is unlikely to happen. I suggest you first try to get familiar with what the library exactly does on `--help`, maybe use a few tools that are based on it, and then re-read carefully this whole discussion from the beginning where UX justifications are provided for that behaviour. 

I guess I should turn that into a cohesive FAQ at some point.

> Actually, it’s `git diff` I’m thinking of - if the diff is a smaller than the terminal, then the pager exits and you don’t have to quit it yourself (which is where this all started!)

Ah that's a different thing from colors of course. Yes that's nice and `FRX` solve that problem. If you know a non-brittle way of making that happen only via `Sys`, I'd welcome a patch.

Issue #27, opened 2015-04-01
UTF-8 manpages


edwintorok:
"However at the moment Groff (at least 1.19.2) doesn't seem to cope with UTF-8 input and UTF-8 characters beyond the ASCII set will look garbled."

I found a workaround: use `preconv(1)` or `groff -k -man -Tutf8`.
Suggestion:
1. Change `Pipe this output to groff -man -Tutf8 | less` to `Pipe this output to groff  -K utf8 -man -Tutf8 | less -s`
2. When automatically running groff, use `-K utf8`
3. `--help=groff` could perform the conversion done by `preconv(1)` on its own, so no `-K utf8` is needed anymore

For example you can currently do this on Debian jessie to get the UTF8 printed correctly:
```sh
./noop --help=groff | groff -k -man -Tutf8 | pager -s
./noop --help=groff | preconv -e utf-8 >noop.1 && groff -man -Tutf8 <noop.1 | pager -s
```

It looks like this happens automatically when you call `man noop` in an UTF-8 locale
(and in a C locale the unicode characters get transliterated to ASCII):
```sh
sudo cp noop.1 /usr/local/man/man1/
man noop
LC_ALL=C man noop
```

The manpage `noop.1` displays even on OpenBSD though not with default flags (by default it replaces all UTF-8 characters with ?):
```sh
mandoc -Tutf8 noop.1 | LC_CTYPE=C.UTF-8 more -s
```

I haven't experienced this behaviour though: "-Tutf8 maps the hyphen-minus U+002D to the minus sign U+2212", and `-Tutf8` is require to get the UTF-8 chars printed at all.

dbuenzli:
Sadly this doesn't seem to work on osx.

dbuenzli:
If I understand correctly `preconv` simply preprocesses the UTF-8 input to `\[uXXXX]`. That's not rocket science but I'm neither willing to add an `uutf` dep on `cmdliner` nor write yet another UTF-8 decoding function.

It's a bit sad `osx` is lagging so much with `groff`. `preconv` was [introduced](http://git.savannah.gnu.org/cgit/groff.git/tree/NEWS#n237) in `groff` 1.20, osx 10.12 has 1.19.2.

dbuenzli:
Prompted by @MisterDA I just had a quick look again at that. 

On macOS 11 the version of groff is still stuck at 1.19.2. However I tested this invocation: 

```
mandoc noop.1 | less
```

[...]

PR #38, opened 2015-11-26
Try man for --help=pager before *roff+$PAGER

drvink:
Add support for "CMDLINER_HELPFMT" and
"CMDLINER_USE_MAN_FOR_ROFF_AND_DISPLAY" environment variables.  The
former takes the same values as "--help=" and if set will be used as its
default argument.  The latter, if present in the environment (its value
is ignored) when producing help for pager mode, will use man for
invoking groff and the pager instead of having cmdliner do it.  This
requires a version of man that will accept a filename argument as the
manpage to process (Linux man-db, FreeBSD man, and OS X man all support
this; possibly others as well).  This also solves, at least for me, the
UTF-8 manpage issue (#27).

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -393,11 +393,11 @@ module Manpage = struct
     match pager with
     | None -> print `Plain ppf v
     | Some pager ->
-        let cmd = match (find_cmd ["groff"; "nroff"]) with
+        let cmds = match (find_cmd ["groff"; "nroff"]) with
         | None ->
             begin match pr_to_temp_file (print `Plain) v with
             | None -> None
-            | Some f -> Some (str "%s < %s" pager f)
+            | Some f -> Some [str "%s < %s" pager f]
             end
         | Some c ->
             begin match pr_to_temp_file (print `Groff) v with
@@ -405,12 +405,17 @@ module Manpage = struct
             | Some f ->
                 (* TODO use -Tutf8, but annoyingly maps U+002D to U+2212. *)
                 let xroff = if c = "groff" then c ^ " -Tascii -P-c" else c in
-                Some (str "%s -man < %s | %s" xroff f pager)
+                Some ["man " ^ f; str "%s -man < %s | %s" xroff f pager]
             end
         in
-        match cmd with
+        match cmds with
         | None -> print `Plain ppf v
-        | Some cmd -> if (Sys.command cmd) <> 0 then print `Plain ppf v
+        | Some cmds ->
+            let rec try_cmd xs = match xs with
+            | [] -> print `Plain ppf v
+            | x::xs -> if (Sys.command x) <> 0 then try_cmd xs
+            in
+            try_cmd cmds
 
   let rec print ?(subst = fun x -> x) fmt ppf page = match fmt with
   | `Pager -> pr_to_pager (print ~subst) ppf page


drvink:
I forgot to mention that this *also* solves the issue of manpage output being wrapped to 80 columns instead of the width of the terminal (again, at least it does for me).

dbuenzli:
There are many problems with this PR. 

1. I'm not really fond of introducing cmdliner specific environment variables. 
2. The use case for `CMDLINER_HELPFMT` is not made at all. In which case did you find the discovered default annoying ? I'd rather improve discovery than have an environment variable for that.
3. The fact that output is wrapped to 80 columns instead of the width of the terminal is a good thing (pickup any good book on typography if you need to be convinced) --- this except if your terminal is narrower, in which case we should format at the given number of columns, however at the moment I don't have a good way of querying the terminal width without introducing further dependencies, something I do not want to do.
4. Your additions are not documented. Neither in the developer documentation nor to the end-user.
5. Your additions do not respect the indentation style of the project as mandated by the `.ocp-indent` file.

Now if invoking `man` directly on the output leads to better results on some platforms, then we could simply try to improve the discovery of such improved `man`s; for sure [POSIX `man`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/man.html) doesn't allow to specifiy the manpage on the command line. 

But then we should not lure people into writing UTF-8 man pages if those are going to be broken on *some* of the platforms. So at the moment effectively looking up for these improved `man`s would only solves 3. but in a bad way: I have the impression that most of the people tend to enlarge their terminal rather than narrow it and since it seems that  `man` programmers are not aware of the fact that stretching lines to an arbitrary extent decreases readability after a certain threshold, it would result in less readable man pages.

drvink:
> I'm not really fond of introducing cmdliner specific environment variables.

Why not? Many libraries provide configuration mechanisms of some sort--for example, readline allows you to configure the behavior of any program that uses readline via the `.inputrc` file. That's more complicated than what's needed here though, because for these options, it is very unlikely that a user would want different behavior for different programs, and it would have taken more code to handle working with a config file compared to simple environment variables, so I thought it best to keep it as simple as possible.

> The use case for `CMDLINER_HELPFMT` is not made at all. In which case did you find the discovered default annoying ? I'd rather improve discovery than have an environment variable for that.

It was relevant while I was testing some changes and I left it in just because I figured someone might want it, but you are right, it is not really necessary. I will remove it.

> The fact that output is wrapped to 80 columns instead of the width of the terminal is a good thing (pickup any good book on typography if you need to be convinced) --- this except if your terminal is narrower, in which case we should format at the given number of columns, however at the moment I don't have a good way of querying the terminal width without introducing further dependencies, something I do not want to do.

I have a copy of _Words Into Type_ a few meters away from me, I am fully aware that there is such a thing as "too wide". :) But consider this: right now, there is no way to get cmdliner to output at anything other than 80 columns. With this patch, if the user chooses to call `man`, that is no longer cmdliner's problem. Furthermore, the user will be able to ask man/groff to wrap at whatever width they want; it is not *forced* to be the width of the terminal. With this patch, any need for extra dependencies is eliminated, users can configure the width to whatever they want, and cmdliner becomes simpler.

> Your additions are not documented. Neither in the developer documentation nor to the end-user.
> Your additions do not respect the indentation style of the project as mandated by the .ocp-indent file.

Oops, these slipped my mind entirely. I will take care of them.

> Now if invoking `man` directly on the output leads to better results on some platforms, then we could simply try to improve the discovery of such improved `man`s

Are you suggesting we should try to detect whether the `man` on the system is capable of calling *roff and the user's pager on its own? I guess if the help mode is `pager`, then we could try calling `man` to see if it knows what to do with a filename, and if it fails, proceed with calling *roff and the pager ourselves, but that seems potentially error-prone (how many `man`s do we want to research to ensure this is robust?) compared to the user just being able to turn it on if they know it will work.

> But then we should not lure people into writing UTF-8 man pages if those are going to be broken on some of the platforms.

I only mentioned the UTF-8 issue because the fact that this fixes it on Linux by happy coincidence was a nice side effect, though if it also fixes it on BSD and OS X, then that would be even better. At any rate, it is unlikely that cmdliner will have much influence on whether people attempt to write UTF-8 manpages or not, so as long as we're not making anything *worse* I don't think there's an issue here.

dbuenzli:
> I'm not really fond of introducing cmdliner specific environment variables.
>> Why not?

Here are a few reasons; because the fact that a program uses cmdliner for its cli should be irrelevant to end-users; because as an end-user it's one more thing I need to configure and tweak when I setup a new environment; because in general defining program behaviour according to environment variables leads to surprises; because introducing configurability instead of doing the right thing is poor design.

> Are you suggesting we should try to detect whether the man on the system is capable of calling *roff and the user's pager on its own?

Yes.

> but that seems potentially error-prone (how many mans do we want to research to ensure this is robust?) 

Yes. That's why I'm a little bit reluctant. Also I didn't have too much complaints so far about the way it works now. So keeping the status quo may be the best thing to do at the moment.

> compared to the user just being able to turn it on if they know it will work.

Most users won't know or care about this and I don't think they should. For this reason I don't think that an env variable is the good mecanism here. Also in the future when ocaml build systems improve I hope we will be able to provide simple push button rules for installing man pages for programs that use cmdliner. 

> At any rate, it is unlikely that cmdliner will have much influence on whether people attempt to write UTF-8 manpages or not, so as long as we're not making anything worse I don't think there's an issue here.

The developer documentation explicitly [tells](http://erratique.ch/software/cmdliner/doc/Cmdliner.html#manual) you not to write UTF-8 man pages and I will not advise otherwise until I have a solution that guarantees me that UTF-8 encoded man pages are readable at least on the three major platforms supported by OCaml.

drvink:
I went for a configuration knob was because it was the simplest approach, and when I don't know what a maintainer prefers, I err on the side of changing as little as possible while still producing something that works, but now I know what you want! I'll change it to attempt to determine whether we can directly invoke `man`, falling back to groff+$PAGER if it fails.

> The developer documentation explicitly tells you not to write UTF-8 man pages and I will not advise otherwise until I have a solution that guarantees me that UTF-8 encoded man pages are readable at least on the three major platforms supported by OCaml.

Oh, I see what you meant now. I wasn't suggesting that you change the docs to say it's ok to write UTF-8 manpages just because this fixes the problem on Linux; rather, that the patch happens to fix it there, and if we're lucky, maybe also BSD and OS X.

drvink:
I've tested that the new changes work on Linux, FreeBSD, and OS X. Since this is now completely transparent to the user, do we still need to note it in the docs? (I ask because there is, for example, no note anywhere in the docs that cmdliner uses specifically `$PAGER` or `$MANPAGER`, so I don't know if this detail is worth mentioning either.)

dbuenzli:
> Since this is now completely transparent to the user, do we still need to note it in the docs?

No.

drvink:
OK. If it looks good to you, it's ready to be merged; if not, please let me know if there's anything else I should fix.

dbuenzli:
The patch is ok. I'll leave this open until the next cmdliner release before merging as I'd like to revisit the UTF-8 problem which may not be solvable by going through `man` directly and also rethink the whole thing with windows in mind once opam on windows lands.

drvink:
Sounds good to me (and it always makes me happy to hear from people who actually do care about Windows users :) ).

dbuenzli:
This does not solve the UTF-8 problem on osx. I'm not sure there's any advantage for this especially since this is a non-standard use of `man` and the current way of doing works well cross platforms.

drvink:
When I tried it, it did solve the issue on OS X. Does it not any longer?

FWIW, the original issue here was that cmdliner only prints at 80 columns and that there is no way at all to configure this; the UTF-8 issue was tangential.

dbuenzli:
> When I tried it, it did solve the issue on OS X. Does it not any longer?

I'm not sure what you tested but given the groff that is being distributed with osx I highly suspect it never did.

drvink:
I no longer have an OS X machine to test on so I can't refute that; either way, it's irrelevant.

Is there some other way we can solve the actual issue here? The short version is that cmdliner does not invoke the "man processing toolchain" (as it were) the same way that most `man`s do, resulting in things looking anywhere from "different" to "wrong" (this is not the only issue I've found in cmdliner, just the only one I could easily diagnose and had time to fix). If you fundamentally disagree then I'll say no more; it's your project and I'm not here to give you a hard time, just trying to help.

dbuenzli:
I have the impression that people rarely run with terminals narrower than 80 columns so it shouldn't be that much of a problem but rather a virtue, see point 3. above.

drvink:
I wrap code and text at 80 columns myself too when writing, but usually when simply reading text I prefer something a bit wider (and we already covered that I am not a stranger either to what makes for pleasant typography or layout). You may recall that some old terminals, including the venerable VT100, had 132 column modes. But no worries. Thanks for at least hearing me out.

PR #53, opened 2016-04-26
Allow setting FMT via Term.info

armish:
Previously, the default FMT was set to `Pager` and there was no easy way to pass another format via `Term.info`. The workaround for this problem (e.g. `darcs` command example) involved re-implementing a print logic via `Manpage.print`, which was not ideal and was introducing an unnecessary amount of boiler-plate code.

This allows setting the `fmt` via `Term.info` as a quick way to customize the default pager option.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -237,6 +237,7 @@ type man_block = [                                 (* block of manpage text. *)
 type term_info =
   { name : string;                                    (* name of the term. *)
     version : string option;                   (* version (for --version). *)
+    fmt: [`Pager | `Plain | `Groff];      (* man page format (for --help). *)
     tdoc : string;                        (* one line description of term. *)
     tdocs : string;       (* title of man section where listed (commands). *)
     sdocs : string;    (* standard options, title of section where listed. *)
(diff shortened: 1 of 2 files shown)


dbuenzli:
I don't understand what your use case is. In any case I think it's better if all programs that use `cmdliner` behave the same way w.r.t. help requests.

dbuenzli:
(And I don't think this has the ability to simplify the `darcs_ex` example in any way)

armish:
sorry for the ambiguous language up there. Let me explain why this PR exists in a clearer way: I have been building a command line utility that accepts multiple arguments/options and I am using your library to handle these. I, however, want the help documentation to be printed in plain text by default; because that is how the rest of the command line tools in my field behave. Currently, to do this, I either have to explicitly suggest user to either:

```
$ export PAGER=cat
# or
$ my_tool --help=plain
```

and I don't want to do this and override the default setting without complicating the code that much. With this change, I can accomplish this via:

```ocaml
let cmd =
	let doc = "My tool is..." in
	let version = "0.0.0" in
	let fmt = `Plain in
	let man = [
		`S "Description";
		`P "$(tname) does this and that.";
	] in
	Term.(const my_tool $ input_file $ output_file),
	Term.(info "my_tool" ~version ~doc ~man ~fmt)

let () = 
	match Cmdliner.Term.eval cmd with 
	| `Error _ -> exit 1
	| _ -> exit 0
```

The only documentation I was able to find about how to override this behavour existed within `darcs_ex`. So this change has nothing to do with simplifying that code, but is meant to allow additional customization regarding how the help text is printed.

Let me know if that makes sense.

dbuenzli:
> The only documentation I was able to find about how to override this behavour existed within darcs_ex.

`darcs_ex` does not this. It has a command `help`, whose output itself, when invoked, is a man page and can be formatted as `--help` allows. It doesn't change the default behaviour. 

Regarding the inability to change the default it's by design. See https://github.com/dbuenzli/cmdliner/pull/26 for a discussion.

armish:
Hmm, completely missed that duplicate PR #26. Looks like we have different opinions about implementing the ability to configure some of these default parameters. I will merge this one into my own fork for my own uses and change the documentation accordingly.

PR #54, opened 2016-05-09
Remove unusued 'default' arg in eval_choice documentation

rleonid:


dbuenzli:
Thanks your patch is in as 44dda8c0002072d

PR #59, opened 2016-07-19
fix typo in documentation: `Arg.env` should be `Arg.env_var`

avsm:


dbuenzli:
Thanks, your patch is in as e5a5a9e8fdfc81.

PR #60, opened 2016-09-08
Fix typos

emillon:
I came across some typos in the source code and documentation, here is a fix.

Thanks!


dbuenzli:
Thanks, your patch is in as  2a0714b9e485173fd001

PR #61, opened 2016-10-02
Fix plain help formatting for commands with empty description.

mstarzyk:
This patch fixes a bug in plain-format help formatting.
When a command has empty description, then a pretty-print box is left open. This breaks all further output.


dbuenzli:
Thanks, your patch is in as 649a6359ea90f2716be

PR #63, opened 2016-11-05
Make the build-system safe-string ready

jpdeplaix:
This is helpful to be able to compile cmdliner with OCaml 4.04.0+safe-string (-safe-string enabled at configure time).

Could you also apply this change to all of your projects that have the same problem ?

### DIFF
--- a/pkg/topkg-ext.ml
+++ b/pkg/topkg-ext.ml
@@ -39,14 +39,18 @@ end = struct
   let read file = try
     let ic = open_in file in 
     let len = in_channel_length ic in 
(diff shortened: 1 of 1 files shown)


jpdeplaix:
Up :)

dbuenzli:
I'm working on gradually making all my packages compile with safe-string to be ready when this become the default. However doing this and making new releases for each for them takes time and making releases only for supporting the `+safe-string` switch is not my biggest priority. On that particular PR, `cmdliner`'s next release will anyways switch to `topkg` the library.

Issue #86, opened 2017-07-24
Implement a topkg and .install free build and install procedure


dbuenzli:
This is needed to bootstrap `opam` on debian.

/cc @hendriktews

dbuenzli:
@hendriktews, I did something in 58e5b06bfec17bc649995b785b4017dbd1a971d8. 

Could you please confirm me that this works for you/is enough (e.g. you can do `cd /tmp && opam source --dev cmdliner`). See the `Makefile`.

hendriktews:
Great. From my point of view you can close this issue. It would be helpful if you could release this, because I could then update the Debian cmdliner package.

dbuenzli:
Thanks  ! See https://github.com/ocaml/opam-repository/pull/10009 or http://erratique.ch/software/cmdliner if you only need the tarball.

PR #87, opened 2017-07-27
add configure to support bytecode systems, DESTDIR and install prefix

hendriktews:
Hi Daniel,

I expanded your topkg free build and install patch a little bit. Your version could be used, but has the following problems:
- no proper DESTDIR support
- no default install prefix
- limited support for bytecode architectures
- fails if installation directories do not exist

This is my proposal, let me know what you think.

Hendrik


dbuenzli:
I'm not going to add a 150 lines `configure` shell script, use `sed` and and whatever for something that simple. A single `Makefile` should perfectly be able to handle this.

> no proper DESTDIR support

If the `DESTDIR` support is wrongly done I'm my proposal please adjust it in the original `Makefile`.

> no default install prefix

I'm not sure exactly why you need one but this can be adjusted in the original `Makefile`

> limited support for bytecode architectures

I'm not sure what you mean here. The appropriate targets are here you can invoke them according to your wishes.

> fails if installation directories do not exist

I see you now use `install` everywhere I didn't do this because on osx this installs only executables I guess this fixes it so why not.

hendriktews:
Daniel Bünzli <notifications@github.com> writes:

> I'm not going to add a 150 lines configure shell script, use sed and
> and whatever for something that simple.

What is the problem? The detection of prerequisites (which takes
most of the 150 lines)? The shell? sed? Do you want everything
just in the makefile?

> I'm not sure what you mean here. The appropriate targets are here you

A proper installation procedure detects itself what to compile
and what to install and does not rely on external intelligence.

dbuenzli:
> What is the problem?

Occam's razor is not being applied.

> A proper installation procedure detects itself what to compile
> and what to install and does not rely on external intelligence.

That's a point of view. I personally see little interest in doing so if the tool I'm using do not provide this support themselves in a generic fashion (which unfortunately `make` and `ocamlbuild` do not provide and those are the tools used for now). For this reasons:

1. You are essentially duplicating information that is in the `opam` file and/or in the build system and hardcoding discovery procedures in the package that should be provided by a library of some sort.
2. I think that the need for people to install `cmdliner` without a package manager of some sort is rather low. The package manager is able to provide that intelligence through its package constraints and build system invocation which allows me to circumvent the weakness of the tools in a sustainable fashion maintenance-wise.

I think that people who are constantly trying to tell me that it has no cost to add this ad-hoc script/Makefile here and there for convenience do not understand what it means to try to scale the development and maintenance of software packages.

So please simply alter in the most minimal fashion the `Makefile` to address the valid points you make. Regarding the byte and native you may be able to discover this by simply including `$(ocamlc -where)/Makefile.config`.

dbuenzli:
I want to move forward on this. I think I addressed most of your points in the `Makefile` via 12137454cd135282ae29d20a3723cc7. Tell me in https://github.com/dbuenzli/cmdliner/issues/86 if further adjustements are needed and/or when you want a release.

hendriktews:
> I want to move forward on this. I think I addressed most of your
> points in the Makefile via 1213745.

Thanks! Please see PR #88 - there I added proper support for
DESTDIR.

Hendrik

hendriktews:
Hi Daniel,

> I think that people who are constantly trying to tell me that it has

I would appreciate if you could realize that I was never trying
to tell you something. If you need to make this statement, please
do it in a channel that addresses the right people.

Hendrik

dbuenzli:
> I would appreciate if you could realize that I was never trying to tell you something.

Sorry, I didn't mean to say this.

PR #88, opened 2017-08-03
add proper support for PREFIX and DESTDIR

hendriktews:
Thanks for adding byte-code only support to the makefile.

DESTDIR however was not working as supposed, please refer to https://www.gnu.org/prep/standards/html_node/DESTDIR.html This PR adds proper support for DESTDIR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,14 +5,14 @@
 # Typical usage:
 #
 # make all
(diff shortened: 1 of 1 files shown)


dbuenzli:
Thanks your patch is in as f71046ddc4e63147c71

Issue #101, opened 2018-08-07
Ability to capture the command line argument consumed by a term


diml:
In dune we sometimes print hints suggesting to run another dune command to get more details, such as:

```
$ dune runtest --profile release
<some build error>
Hint: try: dune external-lib-deps --profile release @runtest
```

The hint needs to replicate many of the command line flags that where passed to `dune build` in order to ensure that the behavior is the same, such as `--profile release` in this example. To implement that, we currently reconstruct the command line while parsing the flags that are common to all command. This is tedious and has been the cause of several bugs in the past.

One way to make this simpler would be to provide a function in `Cmdliner.Term` that would capture the part of the command line consumed by a term. For instance:

```
val capture_args : 'a t -> ('a * string list) t
```

Then we could simply wrap our `common` term with this function.

dbuenzli:
Solved in https://github.com/dbuenzli/cmdliner/pull/103 under the final name `Term.with_used_args`. If you can confirm everything's alright I can do a release.

diml:
I had a look and everything looks good to me. A release is welcome :)

PR #103, opened 2018-10-31
Add Term.capture to capture consumed arguments

diml:
This PR implements the idea described in #101.

I added a test for the new feature, it captures the arguments `-a/--aaa`, `-b/--bbb` and positional arguments:

```
$ ./_build/test/test_capture.native -a 
-a
$ ./_build/test/test_capture.native --aaa
--aaa
$ ./_build/test/test_capture.native -b 0 -a
-a -b 0
$ ./_build/test/test_capture.native -b 0 x y z -a
-a -b 0 x y z
$ ./_build/test/test_capture.native -b 0 --other x y z -a
-a -b 0 x y z
```

### DIFF
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -244,6 +244,11 @@ module Term : sig
   (** [choice_names] is a term that evaluates to the names of the terms
       to choose from. *)
 
+  val capture : 'a t -> ('a * string list) t
+  (** [capture t] is a term whose evaluation captures the actual
+      command line arguments passed by the user that are consumed by
+      [t]. *)
+
   (** {1:tinfo Term information}
 
       Term information defines the name and man page of a term.
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -244,10 +244,10 @@ module Term : sig
   (** [choice_names] is a term that evaluates to the names of the terms
       to choose from. *)
 
-  val capture : 'a t -> ('a * string list) t
-  (** [capture t] is a term whose evaluation captures the actual
-      command line arguments passed by the user that are consumed by
-      [t]. *)
+  val with_evaluated_args : 'a t -> ('a * string list) t
+  (** [with_evaluated_args t] is a term whose evaluation returns the
+      arguments evaluated from the command line passed by the user that
+      are consumed by [t]. *)
 
   (** {1:tinfo Term information}
 
--- a/src/cmdliner_cline.mli
+++ b/src/cmdliner_cline.mli
@@ -17,8 +17,8 @@ val create :
 val opt_arg : t -> Cmdliner_info.arg -> (int * string * (string option)) list
 val pos_arg : t -> Cmdliner_info.arg -> string list
 
-(** Actual command line arguments from the command line *)
 val actual_args : t -> Cmdliner_info.arg -> string list
+(** Actual command line arguments from the command line *)
 
 (*---------------------------------------------------------------------------
    Copyright (c) 2011 Daniel C. Bünzli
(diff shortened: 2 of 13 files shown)


src/cmdliner.mli
```diff
@@ -244,6 +244,11 @@ module Term : sig
   (** [choice_names] is a term that evaluates to the names of the terms
       to choose from. *)
 
+  val capture : 'a t -> ('a * string list) t
```
dbuenzli: I would like to find a better name for this. `Term.capture t` is a bit odd to read: we are not capturing the term but the arguments it concretely parsed. 

Here are a few ideas `Term.{captured,parsed,actual,concrete}[_args}`. This might still be confused with accessing the term's actual specification though. Maybe we should throw an `eval` in there. Something like `Term.captured_eval_args`, bof, or maybe `Term.with_evaluated_args` ?

dbuenzli:
Also @diml due to its unique position I maintain multiple build systems for `cmdliner`. I added a `jbuild` file a long time ago (apparently that never made it in a release), would you be kind enough to upgrade the thing to whatever `dune` consumes nowadays ? (Feel free to add it to this PR).

src/cmdliner_cline.mli
```diff
@@ -17,6 +17,9 @@ val create :
 val opt_arg : t -> Cmdliner_info.arg -> (int * string * (string option)) list
 val pos_arg : t -> Cmdliner_info.arg -> string list
 
+(** Actual command line arguments from the command line *)
+val actual_args : t -> Cmdliner_info.arg -> string list
```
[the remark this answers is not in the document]
diml: Sure, done

diml:
Sure, no problem. I also renamed the `opam` file to `cmdliner.opam` as currently dune uses `<pkg>.opam` files to detect what packages are declared. I also updated the opam file to opam 2.0. Should I update the opam file to use the dune build by default? It speeds up the build: 2s vs 10s on my machine.

src/cmdliner.mli
diml: `with_evaluated_args` seems good

diml: I wonder if it doesn't clash with the evaluation of a term though, the documentation looks a bit odd: `... is a term whose evaluation returns the arguments evaluated ...`. Other suggestions: `with_user_specified_args`, `with_args_from_command_line`.

diml:
@dbuenzli did you have chance to look at this? I'm happy to split out the build system changes in another PR if you prefer.

dbuenzli:
Ah no forgot about it sorry. Will try to merge this in the upcoming days.

dbuenzli:
Thanks your patches are in as 96eaaa034ce1 and ddd71ae9b03f. In the end I prefered `Term.with_used_args` over `Term.with_evaluated_args` see 6866e3137fb74098e. 

> Should I update the opam file to use the dune build by default?

In fact I think I will rather update it to use the bootstrap build system to make sure it's properly working well for system packagers.

dbuenzli:
> In fact I think I will rather update it to use the bootstrap build system to make sure it's properly working well for system packagers.

A good idea, you actually introduced a regression in 96eaaa034ce128a1ae373c104150cd31f4fe4097, see eb29b3f17b67a85d.

diml:
Ok, BTW I believe dune is now widely packaged and has no dependencies apart from OCaml itself, so it's generally an acceptable dependency even for system packages.

dbuenzli:
> Ok, BTW I believe dune is now widely packaged and has no dependencies apart from OCaml itself, so it's generally an acceptable dependency even for system packages.

Last time I interacted with people doing system packages they still had strong policies against vendoring, so dune would likely depend on cmdliner (e.g. on debian).

PR #106, opened 2019-04-08
Sort filenames before building

talex5:
It's useful if building the same source code produces the same binaries, so that multiple people get the same result. I found that sometimes building qubes-mirage-firewall does not produce the expected sha256, and the difference in the ELF binary was in cmdliner symbols.

cmdliner's build system processes files in the order in which they are returned by `Sys.readdir`, which does not return them in any particular order. This patch sorts them first.

### DIFF
--- a/build.ml
+++ b/build.ml
@@ -120,6 +120,7 @@ let ocamldep () = really_find_cmd ["ocamldep.opt"; "ocamldep"]
 (* Build *)
 
 let sort_srcs srcs =
+  let srcs = List.sort String.compare srcs in
   read_cmd (ocamldep () :: "-slash" :: "-sort" :: srcs)
   |> String.trim |> cuts ~sep:' '
 


dbuenzli:
Sorry for nitpicking such a trivial matter. But I wonder the sort shouldn't rather happen [here](https://github.com/dbuenzli/cmdliner/blob/7ceabfd28b54372adca0d34aeea657730a7b9a7d/build.ml#L123). In any case could you please typeset the code to make it fit in 80 columns.

talex5:
I don't think that sort is total:

```bash
$ ocamldep -sort src/cmdliner_suggest.ml src/cmdliner_trie.ml
src/cmdliner_suggest.ml src/cmdliner_trie.ml 
$ ocamldep -sort src/cmdliner_trie.ml src/cmdliner_suggest.ml
src/cmdliner_trie.ml src/cmdliner_suggest.ml
```

The effect is easily seen by running `make && sha256sum _build/cmdliner.cma` with and without a `List.rev` on the end.

I've wrapped the line now.

dbuenzli:
@talex5, sure ocamldep performs a stable sort (preserves the order of unordered values) due to side effecting modules, my suggestion was that by doing there a:
```
let srcs = List.sort compare srcs in
```
before `ocamldep`ing we'd be insulated if someone changes the program and `srcs` suddendly come from somewhere else.

talex5:
OK, I've moved it.

dbuenzli:
Thanks your patch is in as 2e395e35a401786857

PR #108, opened 2019-04-25
Allow to stop options parsing after first positional argument

Khady:
`esy` is using a fork of cmdliner with[ two extra commits](https://github.com/esy-ocaml/cmdliner/commits/feat/stop_on_pos) to allow to stop parsing of options once a first posiitonal argument is detected. Which enables commands such as `esy dune build --help` instead of `esy -- dune build --help` to display the help of dune.

`dune` is also interested by this feature, as discussed there: https://github.com/ocaml/dune/issues/2038#issuecomment-486153110

But dune author would like to have this feature upstream. So here is the pull request.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -131,10 +131,10 @@ module Term = struct
             | Ok true -> Some (Error (`Std_version))
             | Error _ as err -> Some err
 
-  let term_eval ~catch ei f args =
+  let term_eval ~catch ~stop_on_pos ei f args =
     let help, version, ei = add_stdopts ei in
     let term_args = Cmdliner_info.(term_args @@ eval_term ei) in
-    let res = match Cmdliner_cline.create term_args args with
+    let res = match Cmdliner_cline.create ~stop_on_pos term_args args with
     | Error (e, cl) ->
         begin match try_eval_stdopts ~catch ei cl help version with
         | Some e -> e
@@ -215,7 +215,8 @@ module Term = struct
     let term = Cmdliner_info.term_add_args ti al in
     let ei = Cmdliner_info.eval ~term ~main:term ~choices:[] ~env in
     let args = remove_exec argv in
-    let ei, res = term_eval ~catch ei f args in
+    let stop_on_pos = Cmdliner_info.term_stop_on_pos term in
+    let ei, res = term_eval ~catch ~stop_on_pos ei f args in
     do_result help_ppf err_ppf ei res
 
   let choose_term main choices = function
@@ -257,8 +258,9 @@ module Term = struct
         let ei = Cmdliner_info.eval ~term:main ~main ~choices ~env in
         Cmdliner_msg.pp_err_usage err_ppf ei ~err; `Error `Parse
     | Ok ((chosen, f), args) ->
+        let stop_on_pos = Cmdliner_info.term_stop_on_pos chosen in
         let ei = Cmdliner_info.eval ~term:chosen ~main ~choices ~env in
-        let ei, res = term_eval ~catch ei f args in
+        let ei, res = term_eval ~stop_on_pos ~catch ei f args in
         do_result help_ppf err_ppf ei res
 
   let eval_peek_opts
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -219,7 +219,7 @@ module Term = struct
     let ei, res = term_eval ~catch ~stop_on_pos ei f args in
     do_result help_ppf err_ppf ei res
 
-  let choose_term main choices = function
+  let choose_term ~main_on_err main choices = function
   | [] -> Ok (main, [])
   | maybe :: args' as args ->
       if String.length maybe > 1 && maybe.[0] = '-' then Ok (main, args) else
@@ -235,25 +235,31 @@ module Term = struct
       match Cmdliner_trie.find index maybe with
       | `Ok choice -> Ok (choice, args')
       | `Not_found ->
-          let all = Cmdliner_trie.ambiguities index "" in
-          let hints = Cmdliner_suggest.value maybe all in
-          Error (Cmdliner_base.err_unknown ~kind:"command" maybe ~hints)
+          if main_on_err
+          then Ok (main, args)
+          else
+            let all = Cmdliner_trie.ambiguities index "" in
+            let hints = Cmdliner_suggest.value maybe all in
+            Error (Cmdliner_base.err_unknown ~kind:"command" maybe ~hints)
       | `Ambiguous ->
-          let ambs = Cmdliner_trie.ambiguities index maybe in
-          let ambs = List.sort compare ambs in
-          Error (Cmdliner_base.err_ambiguous ~kind:"command" maybe ~ambs)
+          if main_on_err
+          then Ok (main, args)
+          else
+            let ambs = Cmdliner_trie.ambiguities index maybe in
+            let ambs = List.sort compare ambs in
+            Error (Cmdliner_base.err_ambiguous ~kind:"command" maybe ~ambs)
 
   let eval_choice
       ?help:(help_ppf = Format.std_formatter)
       ?err:(err_ppf = Format.err_formatter)
-      ?(catch = true) ?(env = env_default) ?(argv = Sys.argv)
+      ?(catch = true) ?(main_on_err = false) ?(env = env_default) ?(argv = Sys.argv)
       main choices =
     let to_term_f ((al, f), ti) = Cmdliner_info.term_add_args ti al, f in
     let choices_f = List.rev_map to_term_f choices in
     let main_f = to_term_f main in
     let choices = List.rev_map fst choices_f in
     let main = fst main_f in
-    match choose_term main_f choices_f (remove_exec argv) with
+    match choose_term ~main_on_err main_f choices_f (remove_exec argv) with
     | Error err ->
         let ei = Cmdliner_info.eval ~term:main ~main ~choices ~env in
         Cmdliner_msg.pp_err_usage err_ppf ei ~err; `Error `Parse
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -301,7 +301,7 @@ module Term : sig
   val info :
     ?man_xrefs:Manpage.xref list -> ?man:Manpage.block list ->
     ?envs:env_info list -> ?exits:exit_info list -> ?sdocs:string ->
-    ?docs:string -> ?doc:string -> ?version:string -> string -> info
+    ?docs:string -> ?doc:string -> ?version:string -> ?stop_on_pos:bool -> string -> info
   (** [info sdocs man docs doc version name] is a term information
       such that:
       {ul
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -370,7 +370,7 @@ module Term : sig
       uses {!Sys.getenv}. *)
 
   val eval_choice :
-    ?help:Format.formatter -> ?err:Format.formatter -> ?catch:bool ->
+    ?help:Format.formatter -> ?err:Format.formatter -> ?catch:bool -> ?main_on_err:bool ->
     ?env:(string -> string option) -> ?argv:string array ->
     'a t * info -> ('a t * info) list -> 'a result
   (** [eval_choice help err catch argv (t,i) choices] is like {!eval}
--- a/src/cmdliner_cline.ml
+++ b/src/cmdliner_cline.ml
@@ -88,7 +88,7 @@ let hint_matching_opt optidx s =
   | true, [] -> [short_opt]
   | true, l -> if List.mem short_opt l then l else short_opt :: l
 
-let parse_opt_args ~peek_opts optidx cl args =
+let parse_opt_args ~peek_opts ~stop_on_pos optidx cl args =
   (* returns an updated [cl] cmdline according to the options found in [args]
      with the trie index [optidx]. Positional arguments are returned in order
      in a list. *)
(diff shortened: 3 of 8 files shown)


dbuenzli:
I have mixed feelings about this. Having dealt in the past with at least one tool whose cli is context sensitive (namely `docker`) I found it more confusing than enlighting. Of course you can argue this is already the case with `cmdliner` when you use sub commands.

The advantage of the ([standard](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html)) `--` token is that besides being standard, it makes it very clear which optional arguments are for the program that is executed and which are for the sub program you want to execute. 

This is not to say that the pattern you propose is unseen in the wild (e.g. `xargs` is one example) but the `--` token has the advantage of clarity and the scheme you propose does break the the `--` standard itself.

I didn't have a look at the implementation for now but a cursory look seems to indicate it adds undocumented features at the API which of course will need to be solved. Besides, I don't have cmdliner in my head at the moment but I'm not sure about the implementation approach, the fact that you need to specify something at the *term information* level seems suspicious. Also the `main_on_err` seems unrelated to your request or is it ?

Khady:
@andreypopp is the author of the code, so he knows better than me. But I'd be happy to help with the code if possible and if you are open to the idea of this new feature.

andreypopp:
@khady thanks for creating the PR!

@dbuenzli

> I'm not sure about the implementation approach, the fact that you need to specify something at the term information level seems suspicious.

I agree with this. I don't think this should be at term level info but I want this to be configurable per subcommand.

> Also the main_on_err seems unrelated to your request or is it ?

Right, this is an unrelated change. The motivation for this was to enable to have a CLI with subcommands where the default command also can handle any sequence of positonal arguments:

```
% esy install # subcommand
% esy ls --help # default command handles 'ls --help' line
```

Without this change `esy ls --help` will try to lookup a subcommand `ls` and fail.

---

In general I agree with you that such CLI could be confusing:

```
esy --package @opam/cmdliner ls
```

(run command `ls` within the env of `@opam/cmdliner` package)

But in most cases it's just:

```
esy dune build --something
```

Which definitely feels better without the `--` especially given that users will use such invocations often.

Some other popular tools like yarn are already settled on using such pattern and I've never heard someone complaining about it being confusing (just the anecdata).

---

I'd be happy to help fix the implementation if you are open to such changes (or a subset of those).

dbuenzli:
> Which definitely feels better without the `--` especially given that users will use such invocations often.

I think it depends a bit whether these invocations are meant to be read or typed. Once you get to
```
esy --package @thatpkg dune exec --context ctx myexe --opt
```
I would say that `--` helps quite a bit figuring out what is going on.
```
esy --package @thatpkg -- dune exec --context ctx -- myexe --opt
```
also it doesn't seem to me that `--` is that horrible to type. 

In other words I would say that the `--` standard seems to scale well from a readability point of view and it's not too bad from the input point of view.

andreypopp:
> I think it depends a bit whether these invocations are meant to be read or typed

I agree in general, but with esy we are optimising for typing where we usually don't pass options to esy itself, like this:
```
% esy dune build
```

This is because with esy you need to prefix every command with `esy ...`.

Khady:
> esy --package @thatpkg dune exec --context ctx myexe --opt

I think that the use case for this feature is not in this kind of complicated context.
opam exec dune build --profile=release
esy dune build --profile=release
dune exec ./main.bc --main-arg=example
And in my experience with opam and dune, it is very common that a user forgets to add -- at the good place.
Also, this system doesn’t remove the possibility to use --. For example, esy --help -- make, works as expected. So it simplifies commands where a clear separation is not strictly necessary and ease the life of people not very used to cmdliner behavior. While preserving the possibility to have a clear separation between each part.

dbuenzli:
> I think that the use case for this feature is not in this kind of complicated context

Well it seems people do already invoke `dune` via `esy` so you already have two layers of "please pass these arguments to the next program". It's better not to presume too much in which contexts your tools end up being used.

> Also, this system doesn’t remove the possibility to use --. For example, esy --help -- make, works as expected.

If I understood correctly what is being requested here, namely that as soon as you hit the first positional argument all remaining ones are positional, then no you can't. 

I think you are being tricked here in that help commands are treated specially by cmdliner so that you can have parse errors on your cli invocation and still request for help. I don't have it installed to check but I suspect if you try `esy -- make` it will fail trying to execute `--`. 

In general with this convention: 
```
exe posarg -- a0 a1...
```
results in the list of positional arguments `"posarg"`, `"--"`, `"a0"`, `"a1"`, .... But with the `--` convention you get the arguments `"posarg"`, `"a0"`, `"a1"`...

dbuenzli:
Here are a few more arguments that tend to indicate me that this should be supported and not be done by your tools.

1. The scheme only works if you can guarantee that your first positional argument doesn't start with a dash. Not that I saw it in the wild but it's not impossible for a tool name to start with a dash, in which case you are screwed with this proposal. I think this shows that this proposal is somehow a bit unclean and ad-hoc.

2. If you invoke tools that do follow the `--` standard (and this is most of your basic tools, since GNU getopt supports the convention) reading the command lines becomes confusing to a trained cli user. For example if I type or read `esy ls -- -a` I would think I'm giving to `esy` `ls -a` but that's not what happens with this proposal, what is given to `esy` is `ls -- -a` which is confusing.

3. I have at least one user of `dune exec` that mentioned to me that he uses two levels of dashes, e.g. `dune exec x/bench.exe -- --filter crap -- --samples` which helps quite a bit for tracking the levels (note also the second level has no non-dashes positional arguments so you are forced to use the `--` convention there). He's also quite happy by the fact he doesn't have to think too hard when he wants to add arguments to `dune exec`, the `--` makes a clear marker were you need to edit the cli. E.g. moving from `dune exec x/bench.exe -- -fcrap -wt3 -- -ptime` to `dune exec x/bench.exe --p rel -- -fcrap -wt3 -- -ptime`.

dbuenzli:
Someone mentioned to me that point 1. in my last message seems a bit hair pulled so I'd like to clarify it. The thing is that tools in general cannot follow the proposed convention because:

a. Their first positional may really start with a dash (e.g. `printf -- -a`) so in this case you really need the `--` convention.

b. Always following the convention is anyways a no go since it doesn't allow to mix positional and optional argument which is the natural way cli tool work.

Now allowing some of the commands of a tool not to follow the `--` convention is not only inconsistent for the end-user of the tool it also entails that the user needs to have in her mind which tools (or subcommands) follow the convention `--` and which ones do not in order to understand her cli invocations (see point 2. of my previous message for an example) which is a cognitive burden that should not be inflicted on the end user in my opinion. If everyone follows the convention (and again a lot of system tools do) then there is no ambiguity as to how to understand the command lines.

pqwy:
I have a feeling that this would either multiply the rules one needs to keep in mind while either reading or writing command lines, or else unduly constrain them.

The first case happens if one adopts the rule "positional positional arguments stop option parsing for certain subcommands." Assuming that not _all_ subcommands use the rule, the user needs to memorize that, e.g. `dune --p rel utop` and `dune utop --p rel` are both legal, while `dune --p rel exec x.exe` is, `dune exec --p rel x.exe` probably is, and `dune exec x.exe --p rel` is not (or at least means something completely different). As a user, I would rather not keep these rules in my head while trying to get my work done.

The second case happens if one adopts the rule "positional arguments stop option parsing for *all* subcommands," hoping to solve the above. It prohibits e.g. `dune utop --p rel` to bring it in line with `dune exec`. This makes editing command history a mess, something I very often do, and it _still_ adds to the command line rules to remember (which options go on which side of positionals).

And lastly, both cases break clean compositionality of command lines, especially for build tools. I use command cascades with both `topkg` and `dune` -- like `dune exec bench.exe --p rel -- -wt2 -fbleh -- -sctime` -- and the current situation is precise and predictable. In contrast, `dune --p rel exec bench.exe -wt2 -fbleh -- sctime` needs memorizing which program uses which rule.

Here I'm mostly speaking about `dune` and `topkg`. I cannot comment on `esy` -- for all that I know, it might be worthwhile to keep the patched `cmdliner` there.

diml:
It seems to me that the proposed change is already an implicit convention in tools that act as wrapper around another command. Especially in system tools. We never use `--` when we use `ssh` or `sudo` for instance. If they were strict about this convention, we would need to use `--` all the time.

I agree that the situation for `dune utop` is not ideal. Something should be improved here, but I'm not too sure how to do it at this point.

diml:
I did a bit more research on this as I'm interested in this feature for Dune. I looked at commands that take another command as argument, as these behave differently as other commands. I found two conventions:

### Convention 1

The program name stops options of the main program, and `--` is optionally allowed. I.e. the two following pattern are allowed:

- `[main program] [options of main program] [program name not starting with -] [option of sub program]`
- `[main program] [options of main program] -- [any program name] [option of sub program]`

For instance:

```
$ sudo echo -- toto
-- toto
$ sudo -- echo -- toto
-- toto
```

This seems to be the convention for BSD tools such as `ssh` or `sudo`, but also for a general commands such as `xargs`.

# Convention 2

The program name stops options of the main program and command starting with `-` are not allowed. I.e. only the following form is allowed:

```
[main program] [options of main program] [program name not starting with -] [option of sub program]
```

For instance:

```
# chroot / echo -- toto
-- toto
# chroot / -- echo -- toto
chroot: failed to run command ‘--’: No such file or directory
```

This seems to be the convention for tools that use the GNU style of command line options such as `git` (with `git bisect run`) or `chroot`.

# Conclusion

Not all tools behave exactly the same regarding `--`, however none of the tool I tested allows to pass options to the main program after the sub program name. I.e. this is pattern is never allowed:

```
[main program] [program name not starting with -] [options of main program] -- [option of sub program]`
```

So I would say that dune is the program that is not respecting the convention by allowing this last pattern.

@dbuenzli, does that change your view on the current PR? :)

dbuenzli:
> It seems to me that the proposed change is already an implicit convention in tools that act as wrapper around another command.

One difference though is that these commands do not have subcommands. I don't think it's good for end user to expose them to different parsing conventions according to which subcommand they are using. 

Both posix and gnu convention try to expose a simple conceptual model as far as command line argument parsing is concerned: arguments are positional except if they start with a dash in which case their are optional arguments. If you happen to have a positional argument that starts with a dash place them after a `--`: those are unconditionally treated as positional arguments.

That's the model cmdliner exposes and the more tools respect this convention the less our command lines and scripts become confusing, *especially* for commands that invoke other commands. It allows to easily track the meta levels and makes very clear and unambiguous which options are given to which program. It is a compositional model and when I read an invocation [like this](https://discuss.ocaml.org/t/jbuilder-utop-from-emacs-facility/2176/5) I'm glad both `opam` and `dune` follow it.

Again once you have it in your fingers and understand the convention I don't think the `--` is a usability issue.

diml:
> One difference though is that these commands do not have subcommands. I don't think it's good for end user to expose them to different parsing conventions according to which subcommand they are using.

What about `git bisect run`?

> Both posix and gnu conventions try to expose a simple conceptual model as far as command line argument parsing is concerned: arguments for the program to invoke are positional except if they start with a dash in which case their are optional arguments. If you happen to have a positional argument that starts with a dash place them after a `--`: those are unconditionally treated as positional arguments.

Hmm, I don't agree with this: you might have elements in your command line that don't start with a dash and are not positional arguments. The only way to understand how to interpret a command is to know the CLI of the particular program you are calling. For instance, if I write:

```
$ dune exec prog --opt blah
```

Then, maybe `blah` is an argument of `prog` or maybe `blah` is the value of `--opt` and is never passed to `prog` . The only way to distinguish between the two is to look at the documentation for `dune exec`. At least, with the convention that `prog` marks the beginning of sub-command, it is clear that everything after `prog` is part of the sub-command.

> That's the model cmdliner exposes and the more tools respect this convention the less our command lines and scripts become confusing, especially for commands that invoke other commands. 
It allows to easily track the meta levels and makes very clear and unambiguous which options are given to which program. It is a compositional model and when I read an invocation like this I'm glad both opam and dune follow it.

I don't disagree, but I feel like we are now talking about a cmdliner convention rather than a general one. Don't you agree?

Regarding `dune utop`, that's a different case. In this context, `utop` is a sub-command of dune rather than a generic program name. I realise that `dune utop` is confusing, though I'm not too sure how to fix that. Maybe we should have a different command, such as `dune wrap utop` so that one would write `dune wrap [dune options] utop [utop options]`, which is much clearer.

> Again once you have it in your fingers and understand the convention I don't think the -- is a usability issue.

It's not just usability. It's confusing to have options after the inner program name that are passed to the outer one. Especially since it is not what most tools seem to do.

dbuenzli:
> What about `git bisect run`?

The command line world is messy isn't it ? Luckily `cmdliner` is here to enforce good practices without you having to think about it.

> Hmm, I don't agree with this: you might have elements in your command line that don't start with a dash and are not positional arguments.

True. But the principles I gave is a good approximation of it; especially if you use `=` or glued arguments in your lines for option arguments. 

> I don't disagree, but I feel like we are now talking about a cmdliner convention rather than a general one. Don't you agree?

No, that's the convention that is mentioned both by posix and gnu and the more we talk about it the more I appreciate how good, clear and compositional it is... It turns out that cmdliner enforces it and again the more cli tools respect it the better.

> It's not just usability. It's confusing to have options after the inner program name that are passed to the outer one. Especially since it is not what most tools seem to do.

I don't find this extremely confusing. First you can opt not to write it so, then even if that happens the only thing you have to understand as a user or reader is:  *arguments to the sub program are passed as positional arguments* (a natural rule since you don't want options to be interpreted by the invoking program). 

Now you have a simple convention that tells you which "option looking" arguments are going to be passed to the outer program and which one the inner program by just glancing the cli from left to right. 

It turns out that this rule is well supported among cli tools and also works equally for the other sub commands of your tools whether they invoke sub programs (in which case it works recursively if the sub programs are well-behaved) or not which is a very good thing for the user --- consistency, least surprise, etc., ...

The problem with this proposal is that it completely breaks the semantics of `--` resulting in confusing command lines both for readers and invokers which no longer understand where the `--` will end up if they use it (see the examples and problem mentioned above). 

Strictly speaking the only argument that was given in favour of this proposal in this discussion is to avoid typing the `--` but given the many drawbacks I don't think that it's enough to break the convention.

Since I can't see a way to recover what I think are excellent properties of the `--` convention using this proposal I'm going to close this.

Thanks all for your time and input.

PR #109, opened 2019-05-24
New v_opt arguments and sequencing of terms.

reubenrowe:
This adds two features to cmdliner, which I feel could be useful.

- A sequential composition combinator and associated operator for terms.
- Combinators for producing optional arguments that generalise the `opt` and `opt_all` combinators analogously to the way that the `vflag` and `vflag_all` combinators generalise the `flag` and `flag_all` combinators. Specifically, they allow different (collections of) option names to convert their arguments using different converters, and also specify different 'tag' values to associate with the arguments.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -18,6 +18,7 @@ module Term = struct
   (* Terms *)
 
   let ( $ ) = app
+  let ( & ) = seq
 
   type 'a ret = [ `Ok of 'a | term_escape ]
 
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -201,6 +201,12 @@ module Term : sig
   val app : ('a -> 'b) t -> 'a t -> 'b t
   (** [app] is {!($)}. *)
 
+  val ( & ) : 'a t -> 'b t -> 'b t
+  (** [v & v'] is a term that evaluates [v] and then evaluates [v']. *)
+
+  val seq : 'a t -> 'b t -> 'b t
+  (** [seq] is {!(&)}. *)
+
   (** {1 Interacting with Cmdliner's evaluation} *)
 
   type 'a ret =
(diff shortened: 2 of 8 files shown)


dbuenzli:
Hello,

> I feel could be useful.

I do not add features on hypotheses. What is your use case exactly and what are you trying to express ?

In general `Cmdliner` tries to provide combinators for simple and established cli paradigms, so a first step would be to point to a few tools that behave that way. Then the fact that you are not able to make it fit the environment variable support should raise some flags. 

If I understood the particular example you gave, it's not a cli that I would like to use. In my book you would define an option or positional argument for the argument to print and then an option for the formatting (long or short), for the latter `vflag` is entirely sufficient *and* could be baked by a environment variable.

In any case if that's for a tool of your own you can always express the kind of constraints you want by lifting more than one option and resolve the logic in the lifted function and error if needed via [Term.ret](https://erratique.ch/software/cmdliner/doc/Cmdliner.Term.html#VALret).

Regarding sequencing of terms, I'm not sure I see the usefulness of the addition (and in particular I would not like to define new operators), in particular it seems to favour side-effecting evaluation, which should rather be avoided. Again what is your use case ?

reubenrowe:
Hi Daniel,

Yes, the use case is my own tool [ROTOR](https://gitlab.com/trustworthy-refactoring/refactorer), which is a refactoring tool for OCaml.

1. Firstly, motivation for the new `v_opt` combinators.

My tool requires to specify on the command line what the codebase is, meaning which files and directories it contains, which PPX preprocessors are applied to each, and what is the name of the library that each is contained in.

My solution for this is to specify these with different command line options:

- `-f` specifies a file.
- `-d` specifies a directory.
- `--ppx` gives a PPX preprocessor that is to be applied to the most recently given file or directory with the `-f` or `-d` option.
- `--lib` gives the library name that is to be applied to the most recently given file or directory with the `-f` or `-d` option.

Thus, I want to be able to invoke my tool as 

```
> rotor rename -d lib/base/shadow-stdlib/src --lib shadow_stdlib -d lib/base/src --lib base -d lib/bin_prot/shape/src --lib bin_shape_lib -d lib/bin_prot/src --lib bin_prot function_to_rename new_name
```

So, a few things:

- I need to be able to grab all of the `-f`, `-d`, `--ppx`, and `--lib` options *in the order they are given*.
- I need to know which options the values come from.
- I want to be able to use different converters (specifically, `Arg.file` for the `-f` options, and `Arg.dir` for the `-d` options).

It was not obvious to me how to do this with the current interface of `cmdliner`.

You can see how I have implemented this with the new combinators I propose in ROTOR's gitlab repository.

https://gitlab.com/trustworthy-refactoring/refactorer/blob/use_cmdliner/src/infrastructure/codebase.ml#L162

2. For the sequencing combinator.

Yes, it is true I can implement something like this using the `app` combinator, which would only be a little more verbose.

The point about side-effects is a valid one. But OCaml is a side-effecting language, so I think OCaml developers are aware of these kinds of things anyway.

In ROTOR, I want to use this to sequence a term that provides common functionality with a bespoke command for each executable.

https://gitlab.com/trustworthy-refactoring/refactorer/blob/use_cmdliner/src/driver/main.ml#L102

This common functionality is the one implemented here.

https://gitlab.com/trustworthy-refactoring/refactorer/blob/use_cmdliner/src/driver/frontend.ml#L90

dbuenzli:
> My tool requires to specify on the command line what the codebase is, meaning which files and directories it contains, which PPX preprocessors are applied to each, and what is the name of the library that each is contained in.

I'm not sure how scalable that is from a cli perspective. Sometimes you are better off with at DSL that you write in a file. Especially for refactoring tools it seems useful to be able to distribute your refactorings around to be applied on code bases.

That being said, I'm not very fond of the "this option argument applies to the previous argument idea". Except for positional arguments it's generally better to keep options context free, especially for *different* option names.

I think there are various other way you could go around this. First I'm not exactly sure why it's necessary for the operator to distinguish between `-d` and `-f`, that the tool should be able to detect. Then you can embed DSLs in the argument themselves those being positional or optional with values. I mean your line could be just a list of positional arguments:

```
rotor rename OBJ... FNAME RNAME
rotor rename FNAME RNAME OBJ...         # Maybe that's better
```

where each `OBJ` could be of the form (just tried to make up stuff quickly maybe there's something better):

```
[ppx:PPX,][lib:LIB,]PATH
```

dbuenzli:
> In ROTOR, I want to use this to sequence a term that provides common functionality with a bespoke command for each executable.

I also usually do that but I witness the effect in the lifted function via a `()` argument.

reubenrowe:
Thanks Daniel,

You have made some good suggestions. I will go with a command line interface along the lines you suggest.

Many thanks.

Reuben

dbuenzli:
Excellent. Cool project and research by the way, I'll have to read these rotor papers.

PR #111, opened 2019-06-26
Escape absent value

dra27:
`opam admin cache --help` includes this snippet:

```
ARGUMENTS
       DIR (absent=~\opam\cache)
           Name of the cache directory to use.
```
but as the backslashes in the default value weren't escaped, that was causing output on stderr.

### DIFF
--- a/src/cmdliner_docgen.ml
+++ b/src/cmdliner_docgen.ml
@@ -146,7 +146,7 @@ let arg_to_man_item ~errs ~subst ~buf a =
   | Cmdliner_info.Val v ->
       match Lazy.force v with
       | "" -> strf "%s" (or_env ~value:false a)
-      | v -> strf "absent=%s%s" v (or_env ~value:true a)
+      | v -> strf "absent=%s%s" (esc v) (or_env ~value:true a)
   in
   let optvopt = match Cmdliner_info.arg_opt_kind a with
   | Cmdliner_info.Opt_vopt v -> strf "default=%s" v


dbuenzli:
Thanks your patch is in as db4d02a9eb47b5c43127a67c

avsm:
Just a headsup that this changeset seems to cause a regression in the mirage test suite with respect to evaluating `$(mname)`.  With this cmdliner app:

```
open Cmdliner
let run _ = ()
let x = Arg.(value & opt string "$(mname)" & info ["arg"])
let t = Term.(const run $ x)
let () = Term.exit @@ Term.eval (t, Term.info "foo")
```

We see this difference with and without this changeset.

```
diff --git a/foo.expected b/foo.diff
index 86a5da7..bcf0f4a 100644
--- a/foo.expected
+++ b/foo.diff
@@ -5,7 +5,7 @@ SYNOPSIS
        foo [OPTION]... 
 
 OPTIONS
-       --arg=VAL (absent=foo)
+       --arg=VAL (absent=$(mname))
 
        --help[=FMT] (default=auto)
            Show this help in format FMT. The value FMT must be one of `auto',
```

cc @samoht

dbuenzli:
> @dra27's interpretation looks correct to me,

I maintain David's interpretation was right but it turned out I was using the interpretation of the documentation language quite a bit via the `none` string argument of `Arg.some` – it actually hadn't occured to me you could use it the way functoria was doing it in the above commit references.

Somehow the mecanism was wrongly being used for two things: automatic rendering of default values via converter printers and more complex human descriptions of what happens on absence. In 90fbd55 I added an entirely separate way of doing the latter via an `absent` doc string to be given to `Arg.info` which overrides automatic absent values renderings. For example:

```diff
> git diff             
--- a/src/brzo.ml
+++ b/src/brzo.ml
@@ -605,9 +605,9 @@ module Cli = struct
     in
     let docv = "FILE" in
     let env = Cmd.Env.info "BRZO_FILE" in
-    let none = "$(b,BRZO) file at the brzo root" in
-    Arg.(value & opt (some ~none fpath) None &
-         info ["brzo-file"] ~env ~doc ~docv ~docs)
+    let absent = "$(b,BRZO) file at the brzo root" in
+    Arg.(value & opt (some fpath) None &
+         info ["brzo-file"] ~absent ~env ~doc ~docv ~docs) 
```

Quite amazed by how many things I got totally wrong in `Cmdliner`.

PR #120, opened 2020-05-22
Polymorphic return types for exit and exit_status.

favonia:
This will facilitate error handling. These functions never return.

### DIFF
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -447,11 +447,11 @@ module Term : sig
   (** [exit_status_of_status_result] is like {!exit_status_of_result}
       except for [`Ok n] where [n] is used as the status exit code. *)
 
-  val exit : ?term_err:int -> 'a result -> unit
+  val exit : ?term_err:int -> 'a result -> 'b
   (** [exit ~term_err r] is
       [Stdlib.exit @@ exit_status_of_result ~term_err r] *)
 
-  val exit_status : ?term_err:int -> int result -> unit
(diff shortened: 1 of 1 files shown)


dbuenzli:
> This will facilitate error handling.

Could you please expand on that.

favonia:
I am trying to write a program which will examine the arguments more carefully after the mechanism in `Arg`. If the arguments do not make sense, I want to reuse the pretty printers in the library (e.g., the `Term.info` we already prepared) to show the help message and exit the program. This change is an attempt to make the hack slightly cleaner, because I want to mimic the recommended coding style (`Term.(exit ...)`) in a context where the return type is not `unit`. The other reason is that the inherent return types of these two functions are already polymorphic.

I am _not_ happy with my current hack, so I am happy to learn the _correct_ way to do this. To put it more generally, the task is to check some constraints beyond the expressivity of the combinators in `Arg`.

dbuenzli:
> The other reason is that the inherent return types of these two functions are already polymorphic.

The problem is that these kind of changes can break code. So I rather not do this for fancy reasons. 


> To put it more generally, the task is to check some constraints beyond the expressivity of the combinators in `Arg`.

The [`cp` example](https://erratique.ch/software/cmdliner/doc/Cmdliner#excp) in the documentation shows how to do this.

dbuenzli:
I'm closing this since I don't think it's a good idea to change this now. Feel free to continue the discussion here or ask questions on https://discuss.ocaml.org/

favonia:
Thank you for your consideration. I can work around it.

PR #122, opened 2020-07-19
Allow arguments to commands in PAGER and MANPAGER.

verbosemode:
The find_cmd function in cmdliner_manpage.ml runs the test command on commands listed in PAGER and MANPAGER.

If those environment variables contain commands with arguments, the test command returns with a non-zero status.

My current MANPAGER variable is set to "nvim +Man!" and hence doesn't get picked up by cmdliner. This is also true for "most -C" or "most -C -z" for example. 

This PR strips all arguments from the command prior to calling "test".

These were the test cases that I ran prior to and after the code changes.

export MANPAGER=""
export MANPAGER="most -C "
export MANPAGER="most -C -z"
export MANPAGER="nvim +Man!"

As a workaround it is currently also possible to just wrap the command in MANPAGER into a shell script.

I'm looking forward to feedback and suggestions.

### DIFF
--- a/src/cmdliner_manpage.ml
+++ b/src/cmdliner_manpage.ml
@@ -437,7 +437,15 @@ let find_cmd cmds =
   | "Win32" -> "where", " NUL"
   | _ -> "type", "/dev/null"
   in
-  let cmd c = Sys.command (strf "%s %s 1>%s 2>%s" test c null null) = 0 in
+  let trim_cmd c =
+    try
+      String.sub c 0 (String.index c ' ')
+    with Not_found -> c
+  in
+  let cmd c =
+    let c = trim_cmd c in
+    Sys.command (strf "%s %s 1>%s 2>%s" test c null null) = 0
+  in
   try Some (List.find cmd cmds) with Not_found -> None
 
 let pp_to_pager print ppf v =


dbuenzli:
[POSIX `man`](https://pubs.opengroup.org/onlinepubs/007908799/xcu/man.html) says that any string acceptable as a command string to `sh -c` is a valid `PAGER`. 

The question is: to what kind of quoting nightmare for `Sys.command` am I now being dragged into ?

dbuenzli:
Also I think I don't mind not checking `MANPAGER` and `PAGER`'s for the existence of the tool it mentions. Making it fail may actually be less confusing for the end user than trying to "fix" what the user wrongly specified with a default. 

However the problem of how to produce [that `cmd`](https://github.com/dbuenzli/cmdliner/blob/db4d02a9eb47b5c43127a67cb121004b03ea3719/src/cmdliner_manpage.ml#L453) given to `Sys.command` in a sane way remains. A bit too lazy to look it up so please comme up with something sensitive and a rationale that explains why/how it works.

verbosemode:
> However the problem of how to produce [that `cmd`](https://github.com/dbuenzli/cmdliner/blob/db4d02a9eb47b5c43127a67cb121004b03ea3719/src/cmdliner_manpage.ml#L453) given to `Sys.command` in a sane way remains. A bit too lazy to look it up so please comme up with something sensitive and a rationale that explains why/how it works.

Thanks for your feedback. I'm digging a bit deeper into Sys.command and see if I can come up with a more appropriate solution.

dbuenzli:
A bit of fun can certainly be found [here](https://github.com/ocaml/ocaml/pull/1492).

verbosemode:
Sorry for not getting back to this issue earlier. The link you've mentioned is quite scary and I don't think we should try to parse the command.

> Also I think I don't mind not checking MANPAGER and PAGER's for the existence of the tool it mentions. Making it fail may actually be less confusing for the end user than trying to "fix" what the user wrongly specified with a default.

This sounds definitely like the best of all options. However I'm still a bit concerned about users opening issues. It is difficult to estimate how many of them have customized their MANPAGER setup.

But since no one raised a similar issue in the past, it might be not worth fixing or changing the behavior of the library. Maybe we should close it and unearth it later if more feedback is incoming. Using a shell script wrapper for MANPAGER works currently for me quite well.

dbuenzli:
@verbosemode while I did not try to extract the executable from the value of `PAGER` and `MANPAGER`, in 9f6cfe4e28eb7bd04954d982b3d44 I changed the executable detection logic to use `command -v`. 

While this is [not standard](https://pubs.opengroup.org/onlinepubs/9699919799//utilities/command.html) (see the second synopsis) my experience on `zsh` and `bash` is that their built-in `command -v` will not choke on additional cli arguments: 

```
> command -v less -R
/usr/bin/less
```

I hope this change might fix it for you.

Issue #24, opened 2015-03-06
Support for nesting commands


Gbury:
Commands can be specified using the `Term.eval_choice` function, however there doesn't seem to be any way to specify sub commands, to achieve results such as `opam pin add`, where `add` is a sub command of `pin`.

More generally, it would be useful to be able to define arbitrary nesting of commands, possible through the use of a function with the same semantics as `Term.eval_choice`, but with the following signature:

``` ocaml
val group : ('a Term.t * Term.info) list -> 'a Term.t
```

dbuenzli:
Something could be done to improve these use cases at least one level further. What I missed at the time is that often tools having commands also have: 
```
tool OBJECT VERB ...
```
synopses.

However from a usability perspective I don't think it is a good thing to allow arbitrary nesting. Besides the bulk of the work here is not to define a grouping mechanism but rather how to handle and present the resulting automatic documentation in a sensible way (e.g. where do you put the documentation of the  options for each subcommand, etc.).

Gbury:
I think it's reasonable to assume that sub commands of the the same command would have some (if not all) options in common. In that case, each command could have its documentation page, printed when using `tool OBJECT --help`, listing its sub commands in a section and options in other sections (maybe with a section for each sub commands with specific options).

c-cube:
I think having a sub-tree of commands would be useful, and w.r.t. documentation it would require, imho, two things:

- in the documentation of any node of the tree, the whole subtree of commands should be mentioned. Currently it's already the case, since the root documentation mentions the available sub-commands; it should also mention sub-sub-commands, and so on, to summarize the structure of the tree.

- in the documentation of any node of the tree, all options that are specific to this node (`program cmd1 cmd2 cmd3 ...`) should be listed, as well as options specific to prefixes of the tree (the options for `program`, for `program cmd1`, and for `program cmd1 cmd2`. This problem is common in the documentation of object-oriented programs, in which the doc of each class lists both the methods of the class and the methods inherited from parents.

Gbury:
One more thing I just noticed and that could be an issue by itself, but is closely related to this issue, is that when you use `Term.eval_choice`, the only synopsis in the main help page is the one where you call one of the commands, while the synopsis of the default term appears to be discarded.
When the default term simply prints the help page, it doesn't really matter, but as soon as the default term has positional arguments, it is troubling that its synopsis doesn't appear.

dbuenzli:
[...]

PR #123, opened 2020-12-01
Nested Commands #24

rgrinberg:
This is my stab at #24. It seems to be feature complete with arbitrary command nesting, see `groups.t` for a demonstration.

I'm thinking of dropping the default choice at the top level of nesting for consistency with subcommands. That would match dune's behavior of `$ dune` summoning the main help page.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -341,7 +341,14 @@ module Term = struct
         Cmdliner_msg.pp_err_usage err_ppf ei ~err_lines:false
           ~err:"this command has subcommands";
         `Error `Parse
-    | Error (`Invalid_command _) -> `Error `Exn
+    | Error (`Invalid_command (maybe, path, _choices)) ->
+        let err = Cmdliner_base.err_unknown ~kind:"command" maybe ~hints:[] in
+        let sibling_terms = List.map snd choices in
+        let ei = Cmdliner_info.eval ~env
+            (Sub_command { term = main ; path ; main ; sibling_terms})
+        in
+        Cmdliner_msg.pp_err_usage err_ppf ei ~err_lines:false ~err;
+        `Error `Parse
     | Ok (((_, f), info), sibling_terms, path, args) ->
         let sibling_terms = List.map snd sibling_terms in
         let ei = Cmdliner_info.eval ~env
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -331,7 +331,7 @@ module Term = struct
         let err = Cmdliner_base.err_no_sub_command in
         let sibling_terms = List.map snd choices in
         let ei = Cmdliner_info.eval ~env
-            (Sub_command { term = main ; path ; main ; sibling_terms}) in
+            (Sub_command { path ; main ; sibling_terms}) in
         let help, version, ei = add_stdopts ei in
         let term_args = Cmdliner_info.(term_args @@ eval_term ei) in
         let args = remove_exec argv in
@@ -348,23 +348,22 @@ module Term = struct
     | Error (`Invalid_command (maybe, path, choices, hints)) ->
         let err = Cmdliner_base.err_unknown ~kind:"command" maybe ~hints in
         let sibling_terms = List.map snd choices in
-        let ei = Cmdliner_info.eval ~env
-            (Sub_command { term = main ; path ; main ; sibling_terms})
+        let ei =
+          Cmdliner_info.eval ~env (Sub_command { path ; main ; sibling_terms})
         in
         Cmdliner_msg.pp_err_usage err_ppf ei ~err_lines:false ~err;
         `Error `Parse
     | Error (`Ambiguous (cmd, path, ambs)) ->
         let err = Cmdliner_base.err_ambiguous ~kind:"command" cmd ~ambs in
         let sibling_terms = List.map snd choices in
-        let ei = Cmdliner_info.eval ~env
-            (Sub_command { term = main ; path ; main ; sibling_terms})
-        in
+        let ei =
+          Cmdliner_info.eval ~env (Sub_command { path ; main ; sibling_terms}) in
         Cmdliner_msg.pp_err_usage err_ppf ei ~err_lines:false ~err;
         `Error `Parse
     | Ok (((_, f), info), sibling_terms, path, args) ->
         let sibling_terms = List.map snd sibling_terms in
         let ei = Cmdliner_info.eval ~env
-            (Sub_command { main ; term = info ; path ; sibling_terms }) in
+            (Sub_command { main ; path ; sibling_terms }) in
         let ei, res = term_eval ~catch ei f args in
         do_result help_ppf err_ppf ei res
   end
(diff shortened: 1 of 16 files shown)


dbuenzli:
Thanks for your work @rgrinberg. I don't think I'll be able to have a look before the beginning of next year. 

> I'm thinking of dropping the default choice at the top level of nesting for consistency with subcommands.

You mean showing the help ? I was thinking a bit about the default behaviour on `tool group` I think it's better to error with a missing command name and the alternatives rather than show the help. 

In particular for when you use `tool` in scripts via e.g. `tool group $(SUB)` it's preferable to get a non-zero exit code and an error message than a zero exit code and a man page in the face.

dbuenzli:
I had a quick look this morning, it's nice that it doesn't look too invasive. If I understand correctly you do not do anything special regarding docs except for listing `COMMANDS` like `Term.eval_choice` did on the main term – that seems reasonable. 

One thing that nagged me while looking at this is: can we have `Term.eval_choice` piggy back on the new groups (maybe with an ad-hoc internal flag to account for differences if there's any) ?  

Besides, I know `Cmdliner`'s API became an incredible mess (I'm often thinking about offering a more streamlined API under a new toplevel name) but a bit of documentation wouldn't hurt. Could you at least provide a first shot at doc strings in the `.mli`s please.

Also I'm quite convinced that a group without the subcommand should error, not show the help.

rgrinberg:
> One thing that nagged me while looking at this is: can we have Term.eval_choice piggy back on the new groups (maybe with an ad-hoc internal flag to account for differences if there's any) ?

Let me give that a try.

>  Could you at least provide a first shot at doc strings in the .mlis please.

Will do. Usually I do this at the end of the code review, to avoid intermediate changes to the docs. But it sounds like you agree with the API.

> Also I'm quite convinced that a group without the subcommand should error, not show the help.

Sounds good to me. It will require us to change `$ dune` to error instead of displaying help, but that's better behavior as you explained. @jeremiedimino do you agree?

jeremiedimino:
What would the error look like?

dbuenzli:
> It will require us to change `$ dune` to error instead of displaying help,

I don't think that's needed, at the toplevel we need to keep the ability to have a default command. That default command could spit out the help.

rgrinberg:
I added documentation to the `Group` module and implemented `eval_choice` on top of it. Omitting a command results in an error.

Seems to be working well.

dbuenzli:
> Omitting a command results in an error.

Sorry I did not look at the details yet but I'm not sure I got that comment. Can we still get a default command if the toplevel sub-command is omitted ? Breaking this is a no-go.

rgrinberg:
>  Can we still get a default command if the toplevel sub-command is omitted ?

Yes, this behaviour is preserved. Only the omission of subcommands at lower levels results in an error. 

The re-implementation of `eval_choice` in this PR should be indistinguishable from the old one. Any differences are unintentional oversights.

dbuenzli:
Thanks ! This looks very good then. 

As I said it may take me a bit of time to merge (ping if it gets too long). But if further tweaks are needed I'll likely make them myself.  

Just one last thing, I noticed some strange stuttering in `darcs.t` with `TERM=dumb` but that may be a bug that was there before.

rgrinberg:
> As I said it may take me a bit of time to merge (ping if it gets too long). But if further tweaks are needed I'll likely make them myself.

Np. We're in no hurry on the dune side.

> Just one last thing, I noticed some strange stuttering in darcs.t with TERM=dumb but that may be a bug that was there before.

Yes, I was meaning to ask about that. This issue is independent of my PR.

rgrinberg:
Gentle ping @dbuenzli

dbuenzli:
This work is not being ignored :-)

src/cmdliner.ml
```diff
@@ -245,32 +245,139 @@ module Term = struct
...
-    let choices_f = List.rev_map to_term_f choices in
+    let main_args = fst main in
     let main_f = to_term_f main in
-    let choices = List.rev_map fst choices_f in
     let main = fst main_f in
-    match choose_term main_f choices_f (remove_exec argv) with
-    | Error err ->
-        let ei = Cmdliner_info.eval ~term:main ~main ~choices ~env in
+    match choose_term (main_args, (fst main_f)) choices_f (remove_exec argv) with
+    | Error (`No_args (path, choices)) ->
```
shonfeder, line 330: Could we make this configurable, so that the user can supply a default behavior if no subcommand is given? We'd need this for  the behavior planned here: https://github.com/ocaml/dune/issues/4367#issuecomment-815056360

dbuenzli: IIRC. I'd rather not. See the discussion starting [here](https://github.com/dbuenzli/cmdliner/issues/24#issuecomment-495198239).

shonfeder: It seems common to have a natural default behavior for sub commands (see `git branch` or `git remote` or `opam switch` etc.) (as you pointed out on linked conversation, in which, it seems, you were undecided at the time). It's also worth noting that @jeremiedimino, who had suggested the limitation, has actually requested precisely this behavior in the linked comment!

I wonder if the seeming recurrence of this pattern might cause a reconsideration on this point? :)

dbuenzli:
On 11 May 2021 at 20:46:44, Rudi Grinberg ***@***.***) wrote:

> It's what I tried initially actually.

Then we are on the right track!

> I probably won't have time either, but I agree that it's the way to go
> if at all possible. Though personally, I did not find it very difficult
> to refactor dune's cli to use the group based api in this PR. Even if it
> is a little clunky.

As discussed elsewhere I'm already not really happy with the current API so I'd rather not make it worse. Especially since we have the opportunity to add a feature while actually simplifying the API (eval_choice is no longer needed).

D

src/cmdliner.ml
dbuenzli: I'm actually reconsidering. 

I think in the general case selecting a nested command named `cmd : string list` should work as follows. Given command line arguments `args : string list` without the exec name:

1. Let `sel : string list` be the list of all arguments of `args` not prefixed by a `-` and occuring before a potential `--`.
2. If the `cmd` is a prefix of `sel` (and there's no longer command matching `sel`) then `path` is the nested command to use. `cmd` is then removed from `args` and the result is parsed using `path`'s term.

This allows to have a default command at any level (if that command wants positional arguments they have to be specified after a `--`). Besides in contrast to the current behaviour for multi commands which requires the first argument of `args` to be the name of the sub command. It allows to specify options before it, as long as those do not used the glued form (like `-o file` or `--output file`). 

The end user can still end up being confused in many ways while refining cli invocations, but that's a bit in the dna of this poor interface medium and the behaviour seems not too hard to explain and understand.

dbuenzli: Of course that doesn't work :-) (e.g. `js_of_ocaml -o bla.js` breaks, it tries to parse `bla.js` as a command). 

Even though I find that annoying, I think we are forced to require that no optional arguments occur before subcommand specifications (as was always the case in `cmdliner`).

dbuenzli:
Something is now available on master. See the comment [here](https://github.com/dbuenzli/cmdliner/issues/24#issuecomment-1015828118) for further details. 

Thanks @rgrinberg for your work. Despite that I'm not merging this it has not been unuseful and helped to get to the current design.

rgrinberg:
Thanks for your hard work. I'll try master with dune and report how it works.

Rudi.

PR #124, opened 2020-12-22
Term.exit and exit_status_of_result require a unit result

talex5:
Previously, `exit` allowed passing any value to it and just ignored it. This leads to various errors going undetected. e.g. this program returns a successful exit status and displays no output, which is surprising:

```ocaml
let revolt () = Error (`Msg "Revolt!")
open Cmdliner
let revolt_t = Term.(const revolt $ const ())
let () = Term.exit @@ Term.eval (revolt_t, Term.info "revolt")
```

It also causes confusion when an argument is missing, e.g.

```ocaml
let chorus count msg =
  for i = 1 to count do print_endline msg done

open Cmdliner

let count =
  let doc = "Repeat the message $(docv) times." in
  Arg.(value & opt int 10 & info ["c"; "count"] ~docv:"COUNT" ~doc)

let chorus_t = Term.(const chorus $ count)

let () = Term.exit @@ Term.eval (chorus_t, Term.info "count")
```

This program compiled without error, but produces no output.

With this change, both programs report errors at compile-time.

### DIFF
--- a/src/cmdliner.ml
+++ b/src/cmdliner.ml
@@ -276,14 +276,14 @@ module Term = struct
   (* Exits *)
 
   let exit_status_of_result ?(term_err = 1) = function
-  | `Ok _ | `Help | `Version -> exit_status_success
+  | `Ok () | `Help | `Version -> exit_status_success
   | `Error `Term -> term_err
   | `Error `Exn -> exit_status_internal_error
   | `Error `Parse -> exit_status_cli_error
 
   let exit_status_of_status_result ?term_err = function
   | `Ok n -> n
-  | r -> exit_status_of_result ?term_err r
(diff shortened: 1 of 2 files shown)


talex5:
Testing this with the packages I have installed in my switch at the moment:
- `alcotest`, `crowbar` and `opam-0install` were previously correct but needed minor updates for this change.
- `hxd` and `duff.0.2` passed errors to cmdliner expecting them to be reported. This PR would have prevented these bugs (duff removed the executables in 0.3).
- `ocurrent` and `ocaml-ci` were incorrectly passing errors to cmdliner, but also logged them (presumably because I'd been confused by this behaviour in the past).

dbuenzli:
@talex5 you also find a bug in one of my programs which was using `Term.exit` instead of `Term.exit_status` :-) Thanks.

PR #129, opened 2021-10-14
Use exception patterns in pattern matching instead of match try

MisterDA:
They're nicer and also tail-recursive.

https://ocaml.org/manual/coreexamples.html#s%3Aexceptions
https://ocaml.org/manual/patterns.html#sss:exception-match


dbuenzli:
Yes exception patterns are nice. But there's no particular need to change this one. Especially *only this one*.

PR #132, opened 2021-11-17
Always use labels (fix warning 6 labels-omitted).

MisterDA:


dbuenzli:
This was now conflicting but thanks. Should be good on master.

PR #137, opened 2022-03-01
doc/tutorial.md: fixed tiny typo

pveber:


dbuenzli:
Thanks. If only someone would take care of https://github.com/ocaml/odoc/pull/303 these things would not happen.

dbuenzli:
Thanks your patch is in as 01f29b1b9f77acf22c1c7 but I amended it to match the code of [`test/revolt.ml`](https://github.com/dbuenzli/cmdliner/blob/master/test/revolt.ml)

pveber:
Great, thanks for taking care of it!

PR #142, opened 2022-03-14
Fix doc typo.

MisterDA:
Small typo that probably came from copy-pasting the documentation,


dbuenzli:
Thanks you patch is in as 955b1bda4833dd33d4ea4c93

PR #149, opened 2022-04-12
Write library artifacts to `%{lib}%/%{name}%` when using Nix

vapourismo:
I am currently writing a tool that lets you convert OPAM packages directly to Nix derivation and subsequently build them there without any intervention.

While doing that I noticed that `cmdliner` does not install its library artifacts into `%{lib}%/%{name}%`.
From what I can tell so far, most build/package systems follow that convention.

Unfortunately in a Nix setting we also rely on that convention for discoverability. 

If you agree with this change, I would also like to patch the opam files for older cmdliner versions in the [opam-repository](https://github.com/ocaml/opam-repository).

### DIFF
--- a/cmdliner.opam
+++ b/cmdliner.opam
@@ -11,8 +11,22 @@ license: ["ISC"]
 tags: ["cli" "system" "declarative" "org:erratique"]
 depends: ["ocaml" {>= "4.08.0"}]
 build: [[ make "all" "PREFIX=%{prefix}%" ]]
-install: [[make "install" "LIBDIR=%{_:lib}%" "DOCDIR=%{_:doc}%"]
-         [make "install-doc" "LIBDIR=%{_:lib}%" "DOCDIR=%{_:doc}%"]]
+install: [
+  [
+    make
+    "install"
+    "LIBDIR=%{_:lib}%" { os-distribution != "nixos"}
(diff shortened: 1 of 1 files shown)


dbuenzli:
> While doing that I noticed that `cmdliner` does not install its library artifacts into `%{lib}%/%{name}%`.

I don't know what you did but that's not the case (an `ls $(opam var cmdliner:lib)` should be able to convince you). 

In fact `%{_:lib}%` is precisely `%{lib}%/%{name}%`…

vapourismo:
You're right, it should. 

I suppose there must be another reason why `cmdliner` is the only of the 3000 odd packages that needs this type of fix. But no worries, I'll work around this on my end.

For reference, there is no `opam` involved. Hence it can't be used to verify.

dbuenzli:
I'm not sure what you mean by "needing that kind of fix". What you proposed is not a fix, it's plain wrong. It installs `cmdliner` in `libdir/cmdliner/cmdliner`. 

I don't know what you use to interpret the `opam` files but it seems to get the `%{_:var}%` syntax wrong. 

Besides `cmdliner`'s Makefile is a debian compatible `Makefile` so I don't think it should pose any problem for packaging systems out there.

vapourismo:
Yes indeed. That's why I fixed it upstream on my end. Thanks for pointing it out :bow:

Issue #148, opened 2022-04-07
Recommended `Cmd.info_name` doesn't exist


waclena:
I'm updating some `Cmdliner` code to the new API and I have a usage of `Term.name` for which the compiler gives the following warning:
```
Alert deprecated: Cmdliner.Term.name
Use Cmd.info_name instead.
```
but as far as I can tell, `Cmd.info_name` doesn't exist, and I can't find an equivalent function.  Am I missing something?

I'm using Cmdliner version 1.1.1 installed via opam.

Thanks!

dbuenzli:
What's your use exactly ? Can you use `Cmd.name` instead ?

waclena:
Daniel Bünzli ***@***.***> writes:

> What's your use exactly ? Can you use `Cmd.name` instead ?

Well, I found the message confusing, but I guess I just reacted too
quickly.  After I finished converting the first of my (several) Cmdliner
programs, I figured out how to recast the code so that I don't need to
take the name of the Cmd.info anyway.  So feel free to close my issue!

Thanks for Cmdliner!

dbuenzli:
> Well, I found the message confusing

It is ! I was just wondering whether it was ok for you to you not to act on `Cmd.info` but `Cmd.t` instead.

waclena:
Daniel Bünzli ***@***.***> writes:

>> Well, I found the message confusing
>
> It is ! I was just wondering whether it was ok for you to you not to act on `Cmd.info` but `Cmd.t` instead.

Actually, after recasting my previous use of eval_choose to use
Cmd.group, I no longer need Term.info or Cmd.info!  All is well!

dbuenzli:
That was fixed in e23f44933786a4a3a47fc1bd3e845049e5b5d81a.

PR #160, opened 2022-09-01
Fix deprecation warning pointing to a non-existing function

kit-ty-kate:
Fixes https://github.com/dbuenzli/cmdliner/issues/148


dbuenzli:
Thanks.

PR #161, opened 2022-09-01
Add Cmd.alias to create aliases of commands

kit-ty-kate:
I'm currently trying to port opam to use the 1.1 API and this would be incredibly useful to have (otherwise we need to carry the old `(info, term)` tuple around like the old API and it's not really pretty)

### DIFF
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -656,6 +656,9 @@ module Cmd : sig
   val name : 'a t -> string
   (** [name c] is the name of [c]. *)
 
+  val alias : info -> 'a t -> 'a t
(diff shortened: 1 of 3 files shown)


dbuenzli:
If that works sufficiently well it doesn't feel egregious. 

Another slight annoyance is to keep a term when you have it both a subcommand and as the default command but I don't have a solution for that (you'd like `'a Cmd.t -> 'a Term.t` but the datastructure is gone).

kit-ty-kate:
> (you'd like `'a Cmd.t -> 'a Term.t` but the datastructure is gone).

yeah that's what i looked for first but then i saw the new `Cmd.t` data structure and realized it wasn't possible

PR #162, opened 2022-09-05
Display subcommand in error messages

emillon:
When there is an error in the command line (such as an unknown option), this displays the command name instead of just the executable name.

This is visible in the darcs example: `./darcs_ex.exe initialize --invalid` would previously display "darcs: unknown option '--invalid'", but now displays "darcs initialize: ...".

This is also consistent with the `pp_try_help` path.

### DIFF
--- a/src/cmdliner_msg.ml
+++ b/src/cmdliner_msg.ml
@@ -71,7 +71,11 @@ let err_cmd_missing ~dom =
 
 (* Other messages *)
 
-let exec_name ei = Cmdliner_info.Cmd.name @@ Cmdliner_info.Eval.main ei
+let cmds ei =
+  let rcmds = Cmdliner_info.Eval.(cmd ei :: parents ei) in
+  List.rev_map Cmdliner_info.Cmd.name rcmds
+
+let exec_name ei = String.concat " " (cmds ei)
 
 let pp_version ppf ei =
   match Cmdliner_info.Cmd.version @@ Cmdliner_info.Eval.main ei with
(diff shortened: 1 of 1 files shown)


dbuenzli:
Did you check a bit with other command line tool with subcommands what they do ? 

The reason why this is so now is that's a message from the program and the current formatting basically follows the [gnu convention](https://www.gnu.org/prep/standards/standards.html#Errors).

Not necessarily against it but the command name is already mentioned twice just below and your scheme leaves less space for the error message to be seen unwrapped.

dbuenzli:
Thanks for the research @Alizter. 

So to sum up `git` is pretty bad and `nix` is a bit inconsistent.

I'd like to keep the executable name prefix because it's  useful to trace the source of errors in logs or script runs (and follows the gnu convention).

I'm a bit less convinced that it's useful to add the subcommand name since it immediately follows in the sequel. I would tend to favour in making the most of the remaining 80 columns for the error message.

dbuenzli:
Here's a concrete example that overflows (of course you will always have examples that do, even with the current scheme).

```
> b0 cmdlet show --h
b0: option '--h' ambiguous and could be either '--hash-fun' or '--help'
Usage: b0 cmdlet show [OPTION]… [CMDLET]…
Try 'b0 cmdlet show --help' or 'b0 --help' for more information.
```

and

```
> b0 cmdlet show --h
b0 cmdlet show: option '--h' ambiguous and could be either '--hash-fun' or '--help'
Usage: b0 cmdlet show [OPTION]… [CMDLET]…
Try 'b0 cmdlet show --help' or 'b0 --help' for more information.
```

It may be a matter of taste but personally I find it easier to get in the first case that there is a problem and that the `b0` tool wants my attention. 

Namely for me `tool:` is easier to spot than `tool cmd sub:` which can be confused with regular program output on a quick scan.

Besides the functions that implement your sub commands are more likely to output errors using the gnu convention (the executable name is easy to reach for, the cmdliner subcommand a little bit less). So always using the program tool name only is likely to be less confusing than having the subcommands only on cli parse errors.

I think I'm gravitating towards a let's not do that…

emillon:
Hi,
I agree that the overflow is not great, and the full command name is printed just below.
In addition, I thought that this was a change caused by the new `Cmd` API, but no, cmdliner never printed the term name.
So yeah, let's not do this. Thanks!

PR #159, opened 2022-08-30
Fix unexpected-docstring warning when building with dune

kit-ty-kate:
```
$ dune build
File "src/cmdliner.mli", line 577, characters 4-57:
577 |     (** The type for environment variable information. *)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error (warning 50 [unexpected-docstring]): ambiguous documentation comment
File "src/cmdliner.mli", lines 803-806, characters 2-78:
803 | ..(** The type for argument converters.
804 | 
805 |       {b Warning.} Do not use directly, use {!val-conv} or {!val-conv'}.
806 |       This type will become abstract in the next major version of cmdliner. *)
Error (warning 50 [unexpected-docstring]): ambiguous documentation comment
```
Tested with OCaml 4.14.0

PR #169, opened 2022-12-15
Fix warnings

shym:
Fix a series of compiler warnings (warnings 27 unused-var-strict and 50 unexpected-docstring)


dbuenzli:
Thanks but these warnings are not enabled by default. Please avoid making pointless PRs, it wastes everyone's time.

dbuenzli:
See also https://github.com/dbuenzli/cmdliner/pull/159

PR #171, opened 2022-12-15
Use the same set of compiler options between dune and ocamlbuild

kit-ty-kate:
As directed in https://github.com/dbuenzli/cmdliner/pull/159#issuecomment-1352973481

### DIFF
--- a/dune
+++ b/dune
@@ -1,0 +1,1 @@
+(env (_ (flags -bin-annot -safe-string))) ; Use the same flags as with ocamlbuild
--- a/src/dune
+++ b/src/dune
@@ -1,4 +1,3 @@
 (library
  (public_name cmdliner)
- (flags :standard -w -3-6-27-32-35)
  (wrapped false))


dune
```diff
@@ -0,0 +1 @@
+(env (_ (flags -bin-annot -safe-string))) ; Use the same flags as with ocamlbuild
```
kit-ty-kate, line 1: side note: `-bin-annot` is optional (as it is always added to the flags) but it reflects ocamlbuild's `_tags` file more strictly if we add it by hand. But it's up to you, whatever you prefer.

dbuenzli: We don't need to reflect ocamlbuild's `_tags` file the default build is `Makefile` based. We just need to reflect the standard OCaml flags.

src/dune
```diff
@@ -1,4 +1,3 @@
 (library
  (public_name cmdliner)
- (flags :standard -w -3-6-27-32-35)
```
dbuenzli, line 3: And that won't do it, or will it ?

dune
dbuenzli: The standard OCaml *warning* flags.

src/dune
kit-ty-kate: why? By default in dune `(flags :standard)` and the `(env ...)` stanza above simply changes the value of `:standard` for everything is the project. So it just works fine.

dune
kit-ty-kate: I don't really understand the request. The standard warning flag in the one used when the ocamlc/ocamlopt are called without setting `-w` so this is exactly what this is doing as it replaces the set of standard flags that dune has as its default to something that **does not** have `-w`

kit-ty-kate: > We don't need to reflect ocamlbuild's `_tags` file the default build is `Makefile` based. We just need to reflect the standard OCaml flags.

in that case to match the flags set in `build.ml` we need:
```suggestion
(env (_ (flags -g -bin-annot -safe-string))) ; Use the same flags as with build.ml
```

dbuenzli:
Thanks the patch is as 25b5c2159a2d10a802. Sorry for the misunderstanding!

Issue #1, opened 2013-06-03
Support for auto-completion


dbuenzli:
It seems a lot of people want cmdliner to help with auto-completion which seems a rather natural request. There are already quite a few things that could help inside (e.g. optional arguments can be specified by a prefix). 
1. How do we invoke the tool in completion mode ? environment variable ? 
2. For open-ended arguments (e.g. opam package name) how do we specify a user-defined function to be called ? Do we extend `'a converter` or do we add (OCaml) optional arguments to command line argument combinators ? The problem is, is there already sufficient context at that point for the thing to be usefull ? 
3. How can we make it shell independent ? 
- [bash completion tutorial](http://www.debian-administration.org/articles/316)
- [opam completion](https://github.com/OCamlPro/opam/blob/master/shell/opam_completion.sh) 
- [genet completion](http://zoggy.github.io/genet/posts/bash-completion.html)

dra27:
A couple of quick thoughts triggered by another discussion. I agree that hidden (though of course documented!) command line switch seems the best bet, but how about making it one that would be mostly invalid for normal command line syntax? So `foo ++cmdliner++auto` or something?
I also agree that as shell-independent as possible would be good, inasmuch as the library user should not have to write any shell code at all, ideally. Given that any non-trivial completion will need to invoke the command at some point, it should be OK that the syntax completion is mostly implemented in the program itself.
Bash-style completion is (probably) the most common, which would guide the kind of input (either via environment or command line) and output which Cmdliner would expect with this command switch, and it should be relatively easy to wrap a bash-style completion interface with a shell-specific stub script for converting to/from other shells' requirements (see, for example, https://github.com/dra27/opam/blob/windows/shell/opam_completion.lua#L106-L111 and https://github.com/dra27/opam/blob/windows/shell/opam_completion.lua#L267-L292 converting a Lua interface into Bash-style to allow line-by-line translation of a bash completion script).
Perhaps the shell-script stubs is also something Cmdliner might generate automatically (as with manpages) - e.g. `foo ++cmdliner++auto --shell=bash` which generates text to put into foo-complete.sh for installing?
It's then "just" a matter of supplying appropriate hooks in the command line itself :smile:

dbuenzli:
> A couple of quick thoughts triggered by another discussion. I agree that hidden (though of course documented!) command line switch seems the best bet, but how about making it one that would be mostly invalid for normal command line syntax? So foo ++cmdliner++auto or something?

[...]

PR #177, opened 2023-04-13
Add --help=commands

Julow:
This adds `--help=commands`, which prints the list of all the sub-commands accepted by the executable.
This list can be used to quickly navigate between the different manpages and can be parsed.

The output on `test/darcs_ex` is:

    darcs
    darcs initialize
    darcs record
    darcs help

Unlike other help formats, it outputs to stdout and use a strict format that can be parsed: newline-separated list.

This is an argument to the `--help` option instead of a new option because it can also be used by humans.
It is not exposed in the library.

The main motivation is to be able to generate odoc pages containing the manpage of each commands. I've tried this in the past for Odoc (https://github.com/ocaml/odoc/pull/903) and I'm considering this for ocamlformat.
What do you think of this ?

### DIFF
--- a/src/cmdliner_arg.ml
+++ b/src/cmdliner_arg.ml
@@ -317,26 +317,35 @@ let man_fmts =
   ["auto", `Auto; "pager", `Pager; "groff", `Groff; "plain", `Plain]
 
 let man_fmt_docv = "FMT"
-let man_fmts_enum = Cmdliner_base.enum man_fmts
-let man_fmts_alts = doc_alts_enum man_fmts
-let man_fmts_doc kind =
-  strf "Show %s in format $(docv). The value $(docv) must be %s. \
-        With $(b,auto), the format is $(b,pager) or $(b,plain) whenever \
-        the $(b,TERM) env var is $(b,dumb) or undefined."
-    kind man_fmts_alts
+
+let mk_man_format man_fmts doc_kind =
+  let man_fmts_enum = Cmdliner_base.enum man_fmts in
+  let man_fmts_alts = doc_alts_enum man_fmts in
+  let man_fmts_doc =
+    strf
+      "Show %s in format $(docv). The value $(docv) must be %s. With \
+       $(b,auto), the format is $(b,pager) or $(b,plain) whenever the \
+       $(b,TERM) env var is $(b,dumb) or undefined. With $(b,commands), \
+       outputs the list of commands, separated by newlines."
+      doc_kind man_fmts_alts
+  in
+  (man_fmts_enum, man_fmts_doc)
 
 let man_format =
-  let doc = man_fmts_doc "output" in
+  let arg, doc = mk_man_format man_fmts "output" in
   let docv = man_fmt_docv in
-  value & opt man_fmts_enum `Pager & info ["man-format"] ~docv ~doc
+  value & opt arg `Pager & info ["man-format"] ~docv ~doc
 
 let stdopt_version ~docs =
   value & flag & info ["version"] ~docs ~doc:"Show version information."
 
+type help_format = [ `Commands | Cmdliner_manpage.format ]
+
 let stdopt_help ~docs =
-  let doc = man_fmts_doc "this help" in
+  let help_fmts = man_fmts @ [ "commands", `Commands ] in
+  let arg, doc = mk_man_format help_fmts "this help" in
   let docv = man_fmt_docv in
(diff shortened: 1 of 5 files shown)


dbuenzli:
> What do you think of this ?

I think that once https://github.com/dbuenzli/cmdliner/issues/1 is solved in some way this becomes redundant so I'd rather not add this. 

I know #1 has been an extreeeeemly long standing problem but I hope to get to it at some point. Last time I looked at it, it eventually seemed in reach (see the last messages in the issue).

Julow:
That's interesting! Though auto completion is intended to complete command lines and communicate with humans. `--help=commands` is simple enough to be reliably parsed by generator tools.

Do you think auto completion could accept queries like "list all the sub-commands" ? For example, a tool that have deep sub commands:

```
prog
prog foo
prog foo bar
```

A sub-command is anything that can accept `--help` and has no other flags/options.

I want to be sure not to get:

```
prog
prog foo
prog --help
prog --version
```

dbuenzli:
> That's interesting! Though auto completion is intended to complete command lines and communicate with humans.

If you ask it to complete the a command starting with the empty string it will give you all the commands.

In general I'm not decided yet whether it would maybe not be better to dump all the cmdliner metadata in a machine readable format like json which can then be massaged at will by any third-party system, including completion. Or try to have subtler integration with completion like I was starting to get to in #1.

dbuenzli:
> I want to be sure not to get:
[...]

I read to fast yes, of course we should think about that.

Leonidas-from-XIV:
> A sub-command is anything that can accept `--help` and has no other flags/options/positional argument.

But wouldn't the command list that you don't want to get a superset of the information you want? You can search for lines with `--help` and then get the subcommands from there. It even guarantees that the command supports `--help`.

Julow:
It would list all options/arguments for every sub-commands ?

```
prog
prog foo
prog foo bar
prog foo bar --help
prog foo bar <other args>
prog foo --help
prog --help
```

This looks like an interesting output that doesn't require parsing JSON (useful in bash), though the syntax might become complex with added metadata like type of argument (directory, enum).

dbuenzli:
Beyond some form of built-in completion that more and more looks like *Duke Nukem Forever*, I think it would be nice to be able to dump all the metadata (including man pages) cmdliner users provide to the library on say `--help=json` for further processing by third-parties.

The problem is that we'd need to devise a format that is reasonably stable and not too tied to current implementation which takes… time.

Julow:
I could use a metadata dump to achieve my usecase, though the added maintenance to cmdliner looks huge. What are the other use cases for a complete metadata dump ?

dbuenzli:
Not sure if it would be a *maintenance* burden. The burden is rather designing and documenting the format in the first place but:

> What are the other use cases for a complete metadata dump ?

Good question :-) I guess you could certainly then relegate the nasty story of generating completion routines for all these crazy shell completion mecanisms to a third party tool. I guess some people would use it to generate docs in different format without having to parse groff (IIRC correctly that's what `opam` actually does 🥺)

And dreams on. If shells could agree on some kind of machine readable documentation format there's all sorts of nice UI tricks you could beyond completion like get the documentation of the option under my cursor etc.

Julow:
My use case is to generate Odoc pages for a tool's CLI. The current PR would help get the list of commands, which would then be called with `--help` appended and concatenated into a .mld page. This leads to a ugly document.
Having complete metadata would allow to implement this in a nicer way, using more of Odoc's features but requiring the maintenance of an other tool.

Perhaps cmdliner could take care of that ? It supports several documentation formats, except OCaml's one. Designing a new `--help=odoc` format would implement my use case without leaking the metadata (except perhaps the list of commands again).

Julow:
Here's how it looks in ocamlformat, which has only one command: https://ocaml.org/p/ocamlformat/latest/doc/manpage_ocamlformat.html

dbuenzli:
> Designing a new `--help=odoc` format

And the next person will want `--help=md` :-) That the reason why I rather output a json which does "leak" all the data in structured format. 

A tool that converts this json to `.mld` could be part of cmdliner itself. But I don't think it's useful to have it in every cmdliner based program. I'd like to avoid cmdliner adding too much bloat to programs (I already find it too large for what it's doing).

The "multiple formats" that are currently in are there to integrate well in the cli infrastructure.

Julow:
Makes sense. In the meantime, let's close this.

Julow:
This is awesome! Thanks a lot :)

PR #182, opened 2023-12-04
Support env parsing for vflag

kohlivarun5:
Currently, `vflag` does not check environment specification for an argument (like `flag` does). 
This change adds env parsing for `vflag`

### DIFF
--- a/src/cmdliner_arg.ml
+++ b/src/cmdliner_arg.ml
@@ -136,7 +136,7 @@ let vflag v l =
           | Ok true ->
             let opt_name = Cmdliner_info.Arg.opt_name_sample a in
             begin match fv with
-            | None -> (aux [@tailcall]) (Some (Cmdliner_info.Arg.opt_name_sample a, v)) rest
+            | None -> aux (Some (Cmdliner_info.Arg.opt_name_sample a, v)) rest
             | Some (g, _) -> failwith (Cmdliner_msg.err_opt_repeated g opt_name)
             end
           | Error `Parse e ->  failwith e end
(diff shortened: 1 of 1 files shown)


dbuenzli:
What is this supposed to do ?

dbuenzli:
So this adds one env variable per case in the list. I'm not sure I'm very happy about the idea. This makes a lot of irrepresentable states legal e.g. for:

```ocaml
Cmdliner.Arg.(vflag 0 [1, info ["one"]; 2, info ["two"]; 3, info ["three"]]);;
```

You would rather like a single variable to define the actual value rather than three mutually exclusive boolean variables.

src/cmdliner_arg.ml
```diff
@@ -124,11 +124,22 @@ let flag_all a =
...
+        | [] ->
+          begin match try_env ei a Cmdliner_base.env_bool_parse ~absent:false with
+          | Ok false ->
+            (* Env variable being false is equivalent to being absent *)
+            aux fv rest
+          | Ok true ->
+            let opt_name = Cmdliner_info.Arg.opt_name_sample a in
+            begin match fv with
+            | None -> aux (Some (Cmdliner_info.Arg.opt_name_sample a, v)) rest
+            | Some (g, _) -> failwith (Cmdliner_msg.err_opt_repeated g opt_name)
```
kohlivarun5, line 140: > So this adds one env variable per case in the list. I'm not sure I'm very happy about the idea. This makes a lot of irrepresentable states legal

Similar to handling of the `--flag` options, which prevent duplication of arguments:
https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner_arg.ml#L135

The environment variables also prevent duplicated options

kohlivarun5 (commented):
The code prevents representation of illegal states (commented inline about prevention of duplicate options)

> You would rather like a single variable to define the actual value rather than three mutually exclusive boolean variables.

Unfortunately the code I am working with is a third-party library that makes use of `vflag`:
https://github.com/ocaml-ppx/ocamlformat/blob/a7f8da9dabd5b52013df93fc928f389cc54b5bc9/lib/bin_conf/Bin_conf.ml#L147

dbuenzli:
> it shouldn't be hard to convince yourself

I hope you did. Personally I'm not convinced by this proposal.

Issue #186, opened 2024-05-31
Export options information


anne-pacalet:
In [another ticket](https://github.com/dbuenzli/cmdliner/issues/1), you said:

> Today while washing the dishes I thought maybe the best way to try to move on this is that when cmdliner is invoked with `--cmdliner` it should simply dump on stdout all the static information it knows about command lines in some form of structured data format like json or sexp.

For documentation purpose, I would love to be able to export everything about the options in a JSON file. Do you still have this plan in mind?

dbuenzli:
Yes, the recent discussion [here](https://github.com/dbuenzli/cmdliner/pull/177)

anne-pacalet:
Excellent! :+1:

PR #187, opened 2024-06-07
Support for shell completion

andreypopp:
This PR adds support for shell completion (as requested by #1).

Opening it early and as a draft to ask for feedback on the approach.

### How it works, the gist

We define a [shell completion protocol][proto] between a cmdliner-based program
and a shell. 

On the shell side we expect to have a completion function which:

1. invokes a cmdliner-based program with a special completion token injected
   into the argv line

2. interprets the output of the program completion and presents it to the user

We add a helper `cmdliner` tool (to be distributed with the library) which is
used to generate shell completion scripts for a cmdliner-based program for
either bash or zsh (as requested).

### How to test

In this PR I've configured dune to build example programs as executables, it's
convenient to have them on `$PATH` for testing, we also want `cmdliner` tool to
be available:
```sh
export PATH="${PWD}/_build/default/example:$PATH"
export PATH="${PWD}/_build/install/default/bin:$PATH"
```

First we need to install completion handlers for our programs. Let's do that
for a few of example programs:
```sh
eval "$(cmdliner completion-script --shell zsh tail_ex.exe)"
eval "$(cmdliner completion-script --shell zsh cp_ex.exe)"
eval "$(cmdliner completion-script --shell zsh darcs_ex.exe)"
```

The supported values for `--shell` are `bash` and `zsh` as for now.

Now I suggest to try the following completions (I'll paste the output I get as
well with `<TAB>` token showing the place where I've requested completions).

Completion for top level options and subcommands:
```
; darcs_ex.exe <TAB>
--debug        -- Give only debug output.
--prehook      -- Specify command to run before this $(mname) command.
--quiet    -q  -- Suppress informational output.
--verbose  -v  -- Give verbose output.
initialize  -- make the current directory a repository
record      -- create a patch from unrecorded changes
help        -- display help about darcs and darcs commands
```

Completion for options of a subcommand:
```
; darcs_ex.exe initialize <TAB>
--debug        -- Give only debug output.
--prehook      -- Specify command to run before this $(mname) command.
--quiet    -q  -- Suppress informational output.
--repodir      -- Run the program in repository directory $(docv).
--verbose  -v  -- Give verbose output.
```

Completion of an option name:
```
; darcs_ex.exe --<TAB>
--debug    -- Give only debug output.
--quiet    -- Suppress informational output.
--verbose  -- Give verbose output.
--prehook  -- Specify command to run before this $(mname) command.
```

Completion of an option value of enum type (see tail_ex.ml):
```
; tail_ex.exe --follow <TAB>
name        -- name
descriptor  -- descriptor
```

Completion of an argument value and option names:
```
; cp_ex.exe <TAB>
--force      -f      -- If a destination file cannot be opened, remove it and try again.
--recursive  -R  -r  -- Copy directories recursively.
--verbose    -v      -- Print file names as they are copied.
B0.ml           _build/         cmdliner.opam   dune            LICENSE.md      pkg/            _tags
bin/            build.ml*       completion/     dune-project    Makefile        README.md       test/
BRZO            CHANGES.md      doc/            example/        _opam/          src/
```

Completion of an argument value (configured to filename completion):
```
; cp_ex.exe ./<TAB>
B0.ml          build.ml*      completion/    dune-project   Makefile       README.md      test/
BRZO           CHANGES.md     doc/           example/       _opam/         src/
_build/        cmdliner.opam  dune           LICENSE.md     pkg/           _tags
```

Completion of only option names:
```
; cp_ex.exe -<TAB>
--force      -f      -- If a destination file cannot be opened, remove it and try again.
--recursive  -R  -r  -- Copy directories recursively.
--verbose    -v      -- Print file names as they are copied.
```

### TODO

- [x] complete subcommands
  - [x] complete parent command options in subcommand position
- [x] complete option names
  - [x] do not suggest completions for command line flags (they have no value)
- [x] allow to specify completions for option values
  - [x] automatically configure completions for enums
- [x] complete values for positional arguments
- [x] support `zsh` shell
  - [x] separate groups for subcommands vs options
  - [x] fix sort order
- [x] support `bash` shell  
- [ ] support `fish` shell
- [ ] figure out how to write tests
- [ ] get rid of custom dune setup (as I understand dune is not the main build system in the project)
- [ ] decide what to do with `eval_peek_opts`
- [ ] process doc directives when emitting completions
- [ ] do not suggets option names in position after `--`

[proto]: https://github.com/ahrefs/cmdliner/blob/completion/completion/PROTOCOL.md


dbuenzli:
Thanks for getting your hands dirty with this @andreypopp. It is *extremely* appreciated.

I won't have the time to look at this closely for some time. But since you seem to be in the mood, early feedback:

1. I'd like to avoid hardcoding shell support in executables (and hence the library) itself. This doesn't really scale. I'd rather have a well designed generic completion mecanism for cmdliner that is drivable by the various shell completion insanities. Then we can devise hopefully generic completion scripts for the myriad of shells that we maintain on the side (versus embedding the support in executables). We can add a `cmdliner` tool distributed with the library if needed to help with installing completions given a shell and an executable name. 

2. I'd like to avoid driving anything via environment variables. Options that start with `--cmdliner` are reserved I guess this should be enough to do the job. I suspect you can look into the short-circuiting behaviour of `--help` for inspiration.

3. There was also the idea of dumping all the cli metadata in machine readable form via `--help=json`. While this wouldn't cater for dynamic completions it's also perhaps something to keep in mind https://github.com/dbuenzli/cmdliner/issues/186

3. Be careful, we need to support 4.08. I saw a `String.starts_with` in passing which is 4.13, you can add anything that may be missing in `Cmdliner_base`

andreypopp:
Thanks for the feedback!

I've pushed few commits to address some of the feedback items and updated the
PR description accordingly.

More specifically:

> 1. I'd like to avoid hardcoding shell support in executables (and hence the
>    library) itself. This doesn't really scale. I'd rather have a well
>    designed generic completion mecanism for cmdliner that is drivable by the
>    various shell completion insanities. Then we can devise hopefully generic
>    completion scripts for the myriad of shells that we maintain on the side
>    (versus embedding the support in executables). We can add a cmdliner tool
>    distributed with the library if needed to help with installing completions
>    given a shell and an executable name.

I've added a `cmdliner` helper tool that either can be used as

    eval "$(cmdliner completion-script --shell zsh tail_ex.exe)"

or to generate completion scripts to be distributed with cmdliner based programs if needed.

> 2. I'd like to avoid driving anything via environment variables. Options that
>    start with --cmdliner are reserved I guess this should be enough to do the
>    job. I suspect you can look into the short-circuiting behaviour of --help
>    for inspiration.

No environment variables are needed now. But the completion mode is activated
with `+cmdliner-complete:PREFIX` token now. Maybe we can treat this as reserved
as well? Activating with `--cmdliner` is possible too but will probably require
some more code to treat this option name specially.

> 3. There was also the idea of dumping all the cli metadata in machine readable
>    form via --help=json. While this wouldn't cater for dynamic completions it's
>    also perhaps something to keep in mind Export options information #186

Noted. But I think this is a separate feature to be honest. For a good
completion UX context sensitive dynamic completion is needed.

> 4. Be careful, we need to support 4.08. I saw a `String.starts_with` in passing
>    which is 4.13, you can add anything that may be missing in `Cmdliner_base`

Done.

src/cmdliner.mli
```diff
@@ -898,7 +898,9 @@ module Arg : sig
 
   val info :
     ?deprecated:string -> ?absent:string -> ?docs:string -> ?docv:string ->
-    ?doc:string -> ?env:Cmd.Env.info -> string list -> info
+    ?doc:string -> ?env:Cmd.Env.info -> 
+    ?complete:[ `Complete_custom of unit -> (string * string) list | `Complete_dir | `Complete_file ] ->
```
dbuenzli, line 902: We will need a proper type name for that. Also I don't think this belongs to `info` it belongs to the `'a Arg.conv` type. E.g. if I use `Arg.file` converter I shouldn't have to repeat `` `Complete_file ``. 

Currently this type is not abstract but it has been scheduled to become [since 2017](https://github.com/dbuenzli/cmdliner/blob/master/CHANGES.md#v100-2017-03-02-la-forclaz-vs). So probably it's now time to act and make `'a conv` into a proper abstract record. 

Does that still fit your approach ?

andreypopp:
> We will need a proper type name for that. Also I don't think this belongs to
> info it belongs to the 'a Arg.conv type. E.g. if I use Arg.file converter I
> shouldn't have to repeat Complete_file, (same for Arg.enum).

> Currently this type is not abstract but it has been scheduled to become since
> 2017. So probably it's now time to act and make 'a conv into a proper
> abstract record.

> Does that still fit your approach ?

I've [made][1] the `_ conv` type abstract.

Then [moved][2] the completion specification to a converter:

- Had to change term type to keep a map from args to completion specs instead of
  a set, so I can retrieve the completion spec later. I've kept the `Arg.Set` name though...

- Instead of exposing completion spec as a separate type, for now, I've opted
  out to add few arguments to `conv`-family of functions. I think it's simpler
  that way. Let me know what you think.

Then there's another [commit][3] which changes the order we process pos args
when parsing a command line: previously it was right-to-left, now it's
left-to-right. This shouldn't affect actual evaluation, as I understand (as
each pos arg looks only at its own range), but affects completion as the
completion token is now attached to the left most pos arg in "ambiguous" cases
like in `cp_ex.ml SRC... DST`.

[1]: https://github.com/dbuenzli/cmdliner/pull/187/commits/faa77300270a7613959e6fe124541f62e5de747c
[2]: https://github.com/dbuenzli/cmdliner/pull/187/commits/5bebf3362c506b5c92d5e88782f647a1abf5a1fd
[3]: https://github.com/dbuenzli/cmdliner/pull/187/commits/d7780c3084be3d8a63d600954eb5cd4abc046374

andreypopp:
There's one other idea, I think, worth exploring. to make the completion
mechanism more useful - an ability to eval terms in a completion context.

That would allow to support cases where completion might depend on some
configuration, for example a database connection string:

    migrate --db "postgres://localhost:5432/mydb" --migration <TAB>

The code would look like:

    let db_arg =
      Arg.(value & opt (some string) None & info ["db"])

    let migration_arg =
      let complete prefix ctx =
        match Cmdliner.Term.try_eval_for_completion ctx db_arg with
        | None -> (* cannot eval db_arg *) []
        | Some db -> Db.get_unapplied_migrations db
      in
      Arg.(value & opt (some string) None & info ~complete ["migration"])

dbuenzli:
This has been lingering for too long. I promise to proritize it starting on the 1rst of march (famous last words).

dbuenzli:
@andreypopp I have squashed and merged your patches as 4 commits, see the references above (you remain of course the author in contrast to what the UI shows here). 

More work is needed down the line (and more work for 2.0 generally), but I'm very happy with what you have provided, it is excellent bootstrapping work. I will keep this PR open and refer to it from massaging commits until fully done. Thank you.

I'm just curious if you could offer a comment for 7469f4c, that I voluntarly kept separate, is it the scripts that expect that ?

dbuenzli:
> I think, worth exploring. to make the completion mechanism more useful - an ability to eval terms in a completion context.

I suspect this may end up being a bit to tricky to get right, e.g. the program may need to have quite a bit of setup to perform before getting to a workable state and I'm not sure packing these in specific terms is workable or desirable and we would be doing that using a command line that is not fully specified…

Extending the protocol to signify to the completion script that it should call a program to get a list of completions, as suggested by Simmo Saan [here](https://discuss.ocaml.org/t/github-gridbugs-dune-completion-scripts-shell-completion-scripts-for-dune/15410/6) seems perhaps more appropriate.

andreypopp:
> Extending the protocol to signify to the completion script that it should call a program to get a list of completions, as suggested by Simmo Saan [here](https://discuss.ocaml.org/t/github-gridbugs-dune-completion-scripts-shell-completion-scripts-for-dune/15410/6) seems perhaps more appropriate.

But to call a program you'd need to provide some context to it, in a form of arguments, again, so you'd need to access command line anyway. Consider the example above:
```
migrate --db "postgres://localhost:5432/mydb" --migration <TAB>
```
You'd need to call `migrate list-migrations --db "postgres://localhost:5432/mydb"` supposedly. This looks like what I've proposed but with extra steps.

andreypopp:
> I'm just curious if you could offer a comment for https://github.com/dbuenzli/cmdliner/commit/7469f4c61e2bfe2830edca31273dc8f8901781ab, that I voluntarly kept separate, is it the scripts that expect that ?

Sorry, I don't remember 100% off the top of my head and I wish I added a comment earlier. 

I think this is related to optional positional arguments, but not sure, I'll have to recheck.

dbuenzli:
> You'd need to call `migrate list-migrations --db "postgres://localhost:5432/mydb"` 
supposedly. This looks like what I've proposed but with extra steps.

You are right. Right after writing the message I thought, ah but this only work e.g. under an implicit configuration file (thinking `opam install`), you may need to carry over some of the cli flags to the call e.g. `opam --switch=thatswitch install`. I need to think more about it. Still extending the protocol for external *contextualized* invocations may be strictly more powerful as it allows you to hook into other tools. 

> I think this is related to optional positional arguments, but not sure, I'll have to recheck.

It's ok I'll look into it. Though I suspect it has the effect of not completing subcommands when you have a group with a default command and the default command has positional arguments (not something that [is recommended](https://github.com/dbuenzli/cmdliner/blob/b0968e3d268b16e23ceaa175a34a43d55aaf5be6/doc/cookbook.mld#L20-L38) but can be useful, e.g. I [have that](https://erratique.ch/software/b0/doc/occasionally.html#run) in `b0`).

Other things I'm happy to have input about it, especially if you know about these horrendous shell completion routines.

1. I think we should make the scripts totally agnostic to the command. See [this comment](https://github.com/dbuenzli/cmdliner/issues/1#issuecomment-947135989) in my 2021 (!) investigation. We should pass the tool name to the completion function and/or perhaps the function gets called by the shell with the tool name ? I'd prefer that to the meta-programming of the completion scripts like is done now. That way we just install cmdliner completion scripts *once* and people using cmdliner can just call, add to their `.shellrc` or install calls to `compdef _cmdliner_generic mytool`.

2. I would like to extend the protocol and API completion specification with a `reset` token. The effect on the completion script should be to bail out and start completion as if it was the beginning of a shell prompt. Do you have any idea if this is something possible ? 

The use case for 2. is for tools that invoke other tools after a `--`, for example `opam exec --`. By indicating a reset completion e.g. `Arg.Completion.reset` on the positional arguments of  `opam exec` the protocol outputs a `reset` token, which the script should take as start over the completion process.

dbuenzli:
> 1. I think we should make the scripts totally agnostic to the command.

Well the current ones already are :-) I'm starting to wonder whether it's actually useful to provide that `cmdliner` tool rather than simply install the scripts at the right locations and instruct people to: `compdef _cmdliner_generic mytool` and/or install these invocations for their tool (well `cmdliner` could spit that out for them).

dbuenzli:
> * [ ]  figure out how to write tests
> * [ ]  get rid of custom dune setup (as I understand dune is not the main build system in the project)
> * [ ]  decide what to do with `eval_peek_opts`
> * [ ]  process doc directives when emitting completions
> * [ ]  do not suggets option names in position after `--`

These things have been done. For the remaining things to iron out I have opened individual issues and I'm going to close this.

This is taking a bit longer than I thought and I have to switch to something else for a while. I hope that at least at the end of may cmdliner 2.0 will be out.

Regarding some elements of the discussion above: 

> I'm starting to wonder whether it's actually useful to provide that `cmdliner` tool rather than simply install the scripts

The `cmdliner` tool is useful for letting cmdliner library users easily install their completion. The whole scheme, as well as configuration instructions for end-users is [documented](https://erratique.ch/software/cmdliner/doc/cli.html#cli_completion) in the command line interface manual of cmdliner. 

> 2\. I would like to extend the protocol and API completion specification with a `reset` token.

This has been done in 64344e72c2e79aabb5 and [early experiments](https://github.com/b0-system/b0/commit/3b9544a54ee134e3d6c35f976be7b89a2982f290) on `b0 scope exec -- ` and `b0 vcs -- ` seems to indicate that it works quite well, except in [certain cases](https://github.com/dbuenzli/cmdliner/issues/217). The `bash` support also [has to be written](https://github.com/dbuenzli/cmdliner/issues/218).

Thanks again for you help.

Issue #204, opened 2025-03-08
`Term.with_used_args` is very buggy


dbuenzli:
[`Term.with_used_args`](https://erratique.ch/software/cmdliner/doc/Cmdliner/Term/index.html#val-with_used_args) which discussed in #101 and implemented in https://github.com/dbuenzli/cmdliner/pull/103 has many bugs. 

1. It reorders positional and optional arguments; that could be fine.
2. It separates values from short option names. This is a *bug*.
3. It seperates values from long option names. That could be fine.
4. It does not report the `--` token. This is a *bug*.

Shown here on the current test exectuble `test_with_used_args`

```
> b0 -- test_with_used_args -- hoho --bbb=bla
hoho --bbb=bla
> b0 -- test_with_used_args hoho -bbla
-b bla hoho
> b0 -- test_with_used_args hoho --bbb=bla
--bbb bla hoho
> b0 -- test_with_used_args -- hoho --bbb=bla
hoho --bbb=bla
```

PR #210, opened 2025-03-25
Allow mixing opt_all and vflag_all

arozovyk:
Hi, 

Currently, we can define multiple `vflags` with `vflag_all` which returns a list
> that contains one corresponding value per occurrence of the flag, in the order found on the command line. 

This PR extends that functionality by allowing `opt_all` and `vflag_all` to be mixed within the same definition.

The new `opt_vflag_all` function returns a list that contains one corresponding value per occurrence of _the flag or the option_, in the order found on the command line. A GADT is used to convert option values (which may have different types) and unify them with the flag values. 

The goal is to provide users with greater flexibility when defining a logic where the order of applying flags and options matter.

Started from https://github.com/ocaml/opam/issues/6387

### DIFF
--- a/src/cmdliner.mli
+++ b/src/cmdliner.mli
@@ -679,6 +679,26 @@ module Arg : sig
   (** The type for argument converters. See the
       {{!predef}predefined converters}. *)
 
+  type 'a econv = Conv : { conv : 'b conv; vopt : 'b option; vconv : ('b -> 'a)} -> 'a econv  
+  (** A type for handling existentially-typed conversions of optional arguments.
+      Hides the intermediate type ['b] used for parsing option values, allowing different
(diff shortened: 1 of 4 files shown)


dbuenzli:
Tip: always first get in touch first on the issue tracker if you want to add new features. It will be a better use of your time.

This PR has many problem w.r.t. to the API (e.g. do not expose concrete types). It also adds a non trivial amount of definitions to an already complex API.

But more importantly that's not something I want to support in Cmdliner. Context sensitive command lines are not very user friendly. If you find yourself defining this kind of behaviour, in my opinion you should rather question the way you are using the command line.

(If you don't mind munging a bit you can also likely use `Term.with_args` but bear in mind that it seems I was a bit careless when I merged that work and the implementation [is buggy](https://github.com/dbuenzli/cmdliner/issues/204))

PR #241, opened 2025-11-07
bash completion: tell completion system when we are using filenames

### LINKED ISSUE #238: File completion more difficult to use than 'native' tools with bash
The completion provided by the `file` Arg is difficult to use in practice when the file I want to complete is not in the current working directory

For example, compare the completion of the [example_rm](https://erratique.ch/software/cmdliner/doc/examples.html#example_rm) and and 'real' rm, when I have a file in foo/delete_me.txt

`$ example_rm ./f<TAB>` yields `$ example_rm ./foo ` (note the space)
`$ rm ./f<TAB>` yields `$ rm ./foo/` 

The later is much more usable, since I can do `./f<TAB>de<TAB>` and get the full path, where as with the current cmdliner behavior I would need to first hit backspace, add a `/`, and then continue with my path

WardBrian:
Closes #238 

With this, tab-completions for files end with a / and no space, allowing easier completion-chaining

### DIFF
--- a/src/tool/bash-completion.sh
+++ b/src/tool/bash-completion.sh
@@ -15,10 +15,12 @@ _cmdliner_generic() {
         read group
       elif [[ $type == "dirs" ]] && (type compopt &> /dev/null); then
         if [[ $prefix != -* ]]; then
+          compopt -o filenames
           COMPREPLY+=( $(compgen -d "$prefix") )
         fi
       elif [[ $type == "files" ]] && (type compopt &> /dev/null); then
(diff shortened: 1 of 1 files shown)


WardBrian:
Ah, I just noticed that these also live as ocaml strings in https://github.com/dbuenzli/cmdliner/blob/master/src/tool/cmdliner_data.ml. Is one of these the source of the other @dbuenzli?

dbuenzli:
Thanks your patch is in as 225262c4b78b

PR #242, opened 2025-11-08
bash completion: include = as part of completion token

### LINKED ISSUE #231: bash completion of long options with `=`
Note these are issues that should be resolved in the [generic completion script](https://github.com/dbuenzli/cmdliner/blob/master/src/tool/bash-completion.sh).

1. ~~`cmdliner install tool-support --shell=␉` gets stuck. This seems to be related to `=` being part of `COMP_WORDBREAKS` but I couldn't figure out how I could work around this.~~ Fixed by [c22bce1](https://github.com/dbuenzli/cmdliner/commit/c22bce1b02d24c1c9f8e77ea82ba6d15699a7495)
3. `cmdliner install tool-support --sharedir=␉` Currently I have no idea how to combine the file/directory completion with glued forms (short forms included)

Note that by default if you tab your way through the command line you hop to a new token after having completed a long option and thus get a non-glued form whose value completion works. So that helps keeping the user experience reasonably good.

WardBrian:
Resolves #231 

`_get_comp_words_by_ref` is a rather mysterious function that is undocumented but seems to be widely used (e.g., it's in the recommended git bash completion script). It allows us to exclude (`-n`) characters (in this case, just `=`) from the word breaking.

### DIFF
--- a/src/tool/bash-completion.sh
+++ b/src/tool/bash-completion.sh
@@ -1,12 +1,7 @@
 _cmdliner_generic() {
   local words cword
-  if (type _get_comp_words_by_ref &> /dev/null); then
-    # Equivalent of COMP_WORDS, COMP_CWORD but allow us to  exclude '=' as a word separator
-    _get_comp_words_by_ref -n = words cword
-  else
-    words=("${COMP_WORDS[@]}")
-    cword=$COMP_CWORD
-  fi
+  # Equivalent of COMP_WORDS, COMP_CWORD but allow us to  exclude '=' as a word separator
+  _get_comp_words_by_ref -n = words cword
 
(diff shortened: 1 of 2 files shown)


WardBrian:
@dbuenzli I cleaned this up a little to avoid failing if the `_get_comp_words_by_ref` function isn't available (it will just degrade back to a version that doesn't complete after `=`)

dbuenzli:
Mmh but I think that whole installing/automatically loading mechanism also depends on it. 

Also I don't want to have cli devs banging their head on the table as to why the restart completions don't work for their users to later find out it's because they users didn't have that package installed. So I think it's better if we communicate that it's required and behave as such.

PR #244, opened 2025-11-11
Support file and directory completions immediately following an option

### LINKED ISSUE #231: bash completion of long options with `=`

WardBrian:
This closes #231 (addresses bullet 2).

You can now do `test_shell --file=<TAB>` and get suggestions. Short options like `-ds<TAB>` also work, although it looks adding back the `-d` we need breaks the feature that ends the completion with `/` for directories, which is a shame (so the previous completes to `-dsrc` in this repo, for example).

### DIFF
--- a/src/tool/bash-completion.sh
+++ b/src/tool/bash-completion.sh
@@ -17,7 +17,7 @@ _cmdliner_generic() {
     while read type; do
       if [[ $type == "group" ]]; then
         read group
-      elif [[ $type == "dirs" ]] && (type compopt &> /dev/null); then
+      elif [[ $type == "dirs" || $type == "files" ]] && (type compopt &> /dev/null); then
         # trim option prefix in cases like --file=<TAB> or -f<TAB>
         local pattern="$prefix"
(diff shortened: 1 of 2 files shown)


dbuenzli:
Thanks!

PR #245, opened 2025-11-11
Powershell completion

### LINKED ISSUE #214: Support for `pwsh`
Follow up to #1 and #187. 

Along [these lines](https://erratique.ch/software/cmdliner/doc/cli.html#cli_completion). This means:

1. End-user configuration instructions.
2. Determine `opam` install location.
3. A generic completion script.

See the directory [src/tool](https://github.com/dbuenzli/cmdliner/tree/master/src/tool) and [these instructions](https://github.com/dbuenzli/cmdliner/blob/master/DEVEL.md#changing-completion-scripts).

WardBrian:
Closes #214 

Open questions: 
- How to test `restart` directives
- What does `sharedir` even mean for this shell/is there a good place to install these

### DIFF
--- a/src/tool/cmdliner_data.ml
+++ b/src/tool/cmdliner_data.ml
@@ -246,8 +246,8 @@ $Global:%s = {
 
         $CompletionResults.Add(
           [System.Management.Automation.CompletionResult]::new(
-            $completionItem,
             $item,
+            $completionItem,
             'ParameterValue',
             $itemDoc))
       }
--- a/src/tool/pwsh-completion.ps1
+++ b/src/tool/pwsh-completion.ps1
@@ -97,8 +97,8 @@ $Global:_cmdliner_generic = {
 
         $CompletionResults.Add(
           [System.Management.Automation.CompletionResult]::new(
-            $completionItem,
             $item,
+            $completionItem,
             'ParameterValue',
             $itemDoc))
       }


dbuenzli:
> * How to test `restart` directives

If you have `b0` working in the powershell you can do `b0 vcs -- git ␉` but I will add a test to `test_shell`.

dbuenzli:
I added a test to `test_shell` in ebea98b5521a3 but note that it seems the raw completion done by `b0` see [here](https://discuss.ocaml.org/t/ann-cmdliner-2-0-0/17324/2#p-72908-compositional-completion-restart-and-raw-2) for an explanation breaks something. So you can't use `b0 -- test_shell` to test it. For example for testing `zsh` I had to do: 

```
compdef _cmdliner_generic test_shell
```
and then invoke the tool via the path output with: 
```
> b0 --path -- test_shell             
/Users/dbuenzli/sync/repos/cmdliner/_b0/b/user/test_shell/test_shell
```

WardBrian:
I think this is basically ready to go in terms of code. The main question I have is in terms of documentation -- using these scripts will be less automatic than for zsh/bash, since each individual tool will need to be sourced in [the $profile file](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.5) I believe, unless we wanted to add some logic to the generic script that would loop over it's current directory and source other files automatically

WardBrian:
tip for anyone who edits the .ps1 script going forward:

if you find that a file completion is being provided unexpected, that probably means that the script failed with an error and the tab completion machinery went back to it's default behavior. 
Errors inside the completion script are not directly reported. You can inspect the most recent N errors with `Get-Error -Newest N`. This would have saved me a lot of time if I had known it at the start of writing this

dbuenzli:
Perhaps worth adding as a comment to the script? 

I will merge all your stuff in the upcoming days. One thing that is missing but I'm ok to do (I might ask you to re-read it) is to have a section about `pwsh` in end user configuration [here](https://github.com/dbuenzli/cmdliner/blob/1fac92adf9ceb6f8929e2fc2864f35d539626838/doc/cli.mld#L194).

dbuenzli:
Thanks your patch is in as 3d8d771bc4644

PR #247, opened 2025-11-14
Document dune installation of tool support

WardBrian:
Note that I also did get the currently-recommended set up to work using a .opam.template file, so this isn't strictly necessary, but I think this is a more dune-friendly way to get these files in your package by directly hooking them into `dune build @install`

If `cmdliner` wanted to be really nice, it could even provide a way of generating this config snippet so it could be used with dune's `dynamic_include`. If that's something you wouldn't mind, happy to PR it.

### DIFF
--- a/doc/cookbook.mld
+++ b/doc/cookbook.mld
@@ -217,19 +217,42 @@ invocation.
 
 First make sure your understand the
 {{!tip_tool_support_with_opam}above basic instructions} for [opam].
-You then
-{{:https://dune.readthedocs.io/en/stable/reference/packages.html#generating-opam-files}need to figure out} how to add the [cmdliner install] instruction to the [build:]
-field of the opam file after your [dune] build instructions. For a tool named
-[tool] the result should eventually look this:
 
-{@sh[
-build: [
-   [ … ] # Your regular dune build instructions
-   ["cmdliner" "install" "tool-support"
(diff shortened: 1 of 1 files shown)


WardBrian:
The closest I've gotten to something that works without #248 is pretty bad:

```
(rule
 (copy stanc.exe stanc))

(rule
 (target
  (dir cmdliner-support))
 (deps stanc)
 (action
  (ignore-outputs
   (run cmdliner install tool-support ./stanc cmdliner-support))))

(install
 (section share_root)
 (files cmdliner-support/share))
```

This is janky because: 

- `cmdliner install tool-support` doesn't like the .exe suffix, so you have to copy your exe
- `dune` sucessfully copies the files to `_build/install/share` and _then_ errors because of `Unexpected file kind "S_DIR" (directory)`


I think the second issue could be fixed if cmdliner generated a list of files that dune was able to consume, for example

`cmdliner install tool-support ./stanc cmdliner-support --dune-outputs=files.sexp`
```
; file files.sexp
(
(cmdliner-support/share/man/man1/stanc.1 as man/man1/stanc.1)
; ...
)
```

The `install` stanza would then just be `(files (include files.sexp))`

This could be another flag to `install` like the `--update-opam-install`, rather than its own command -- does that make it more appealing?

Issue #239, opened 2025-11-07
`list` helper does not provide completion


WardBrian:
I'd expect completion to define a tab completion for each element that is just whatever the underlying converter provides. e.g., `--foo=<TAB>` or `--foo=bar,<TAB>` would both provide suggestions

dbuenzli:
Also for the specific format you mention here (`=`) there is the more general issue https://github.com/dbuenzli/cmdliner/issues/231

dbuenzli:
This is proving a little bit more complicated than I expected because directives are [typed](https://erratique.ch/software/cmdliner/doc/Cmdliner/Arg/Completion/index.html#type-directive), abstract and not introspectable (my attempt below). So it's not possible to transform the result of the base converter, that is the `'a directive list` to a `'a list directive.t list`. I'd say that it's not even always clear how'd you do that, e.g. on file or raw directives, but you could bail out/ignore these cases.

If you need that now I suggest you define your to devise a primitive that you use both in your base `conv` and an ad-hoc `list_conv`.

(More generally while it can have it's uses I think it's better to avoid defining too complex languages in optional values)

```sh
  let completion = match Completion.complete (Conv.completion conv) with
  | Completion.Complete (context, complete) ->
      let complete ctx ~token =
        match List.rev (String.split_on_char sep token) with
        | [] ->
            let directives = complete ctx ~token in
            failwith "TODO need to get to 'a list directive"
        | arg :: args ->
            let directives = complete ctx ~token in
            failwith "TODO need to get to 'a list directive"
      in
      Completion.make ?context complete
  in
  Conv.make ~docv ~parser ~pp ~completion ()
```

WardBrian:
@dbuenzli thoughts on providing the ability to get completion at least on the first element? I don't see any way to implement this outside the library because of the abstract type

dbuenzli:
> (Alternatively, a value_map function `('a -> 'b) -> 'a directive -> 'b directive`? Then I believe I could keep this hack in my own code)

Yes actually that was the angle I was seeking to pursue. I will have a deeper look at that before the next release.

dbuenzli:
> In particular, no `value_map`

Also it's not enough because that would handle only the "typed" case.

dbuenzli:
> At any rate, my original purpose for opening this issue has been resolved, so feel free to close it if you don't think further support within the library itself is reasonable.

[...]

PR #249, opened 2025-11-17
pwsh: quote completion reply when necessary

WardBrian:
When testing my [list completer function](https://github.com/dbuenzli/cmdliner/issues/239#issuecomment-3542936104) I discovered that powershell requires you to quote strings with commas in them if you want them treated as one value, so you'd write something like `mytool --foo="bar,b<TAB>"`, and the reply needs to be quoted as well. 

I suspect there will be more values than just `,` that will need to be tested for in replies, but this is a start.

### DIFF
--- a/src/tool/cmdliner_data.ml
+++ b/src/tool/cmdliner_data.ml
@@ -180,7 +180,8 @@ $Global:%s = {
     if ($otherArgs -ne "") {
       $otherArgs += " "
     }
-    if ($text -eq $wordToComplete) {
+    # wordToComplete has had quoting removed, so we need to remove it before comparison
+    if ($text -replace "[`"']", "" -eq $wordToComplete) {
       $otherArgs += "--__complete=$text"
       $seenWordToComplete = $true
     }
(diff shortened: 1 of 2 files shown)


dbuenzli:
> I suspect there will be more values than just `,` that will need to be tested for in replies, but this is a start.

Can we perhaps have a more principled approach to the problem :–) (though it seems [the documentation](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.completionresult?view=powershellsdk-7.4.0) is poor)

dbuenzli:
> it appears that `,`, `|`, and `;` all do, with the later two usually ending the command entirely.

Technically values can be anything.

> So if we expect completions to return those, they would need to be quoted (I expect the same could be true in the other shells, at least for the pipe?)

Somehow I'd expect the completion engine to be given values and the UI machinery to do the quoting as needed. But I guess I'm dreaming at that point.

dbuenzli:
Thanks! Your patches are in as d4acc611eade3c

PR #260, opened 2026-01-20
Enable ocamlfind lookup of executables

### LINKED ISSUE #259: Cross compilation support?
f5712268ea16e558d4e19ccbaf964157e5fdf766 removed the dune files, which were being used by [opam-cross-windows](https://github.com/ocaml-cross/opam-cross-windows/blob/main/packages/cmdliner-windows/cmdliner-windows.2.1.0/opam#L33). I've tried `OCAMLFIND_TOOLCHAIN=windows make` but this appears to just run the standard build.

WardBrian:
Closes #259 

For something like opam-cross-windows, the following is now sufficient:

```
make build-byte build-native build-native-exe OCAMLFIND_TOOLCHAIN=windows
```

I didn't do all the makefile work necessary to have `build-completions` work, as this requires a host copy of the cmdliner tool, which would mean detecting when cross compilation is happening and building it twice. It's possible, but it doesn't seem that valuable; `opam-cross-windows` is for building binaries for end users, so the completions etc would be provided separately and likely with `--standalone-completions`

### DIFF
--- a/build.ml
+++ b/build.ml
@@ -121,16 +121,16 @@ let rec rmdir dir =
 
 let really_find_cmd alts =
   match Sys.getenv_opt "OCAMLFIND_TOOLCHAIN" with
+  | None | Some "" -> (
+      match find_cmd alts with
+      | Some cmd -> [cmd]
+      | None -> err "No %s found in PATH\n" (List.hd @@ List.rev alts))
   | Some toolchain -> (
       match find_cmd [ "ocamlfind" ] with
       | Some ocamlfind ->
         [ocamlfind; "-toolchain"; toolchain; List.hd @@ List.rev alts]
       | None ->
           err "OCAMLFIND_TOOLCHAIN is set but ocamlfind not found in PATH\n")
-  | None -> (
-      match find_cmd alts with
-      | Some cmd -> [cmd]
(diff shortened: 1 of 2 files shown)


build.ml
```diff
@@ -119,9 +119,18 @@ let rec rmdir dir =
 
 (* Lookup OCaml compilers and ocamldep *)
 
-let really_find_cmd alts = match find_cmd alts with
-| Some cmd -> cmd
-| None -> err "No %s found in PATH\n" (List.hd @@ List.rev alts)
+let really_find_cmd alts =
+  match Sys.getenv_opt "OCAMLFIND_TOOLCHAIN" with
+  | Some toolchain -> (
```
dbuenzli, line 128: Can you just bring `Some ""` into the `None` case please or, alternatively report an error like `err "OCAMLFIND_TOOLCHAIN is set but empty"`.

dbuenzli (changes requested):
Thank you very much. Just one thing to tweak please.

dbuenzli:
Thanks!

dbuenzli:
(and sorry for [breaking your workflow](https://xkcd.com/1172/))

WardBrian:
You know, I've seen that comic dozens of times and this was the first time that the joke of the specific choice of key dawned on me (it's a _space_ heater)

PR #263, opened 2026-02-01
Allow completion even when legacy prefixes lead to ambiguity

### LINKED ISSUE #262: Cmdliner completions break if CMDLINER_LEGACY_PREFIXES is true at --__complete time
For example, this actually works fine:

```bash
$ CMDLINER_LEGACY_PREFIXES=1 stanc -<TAB>
-?                              --debug-generate-inits          --debug-transformed-mir         --info                          --print-cpp
```

But this, which would usually be equivalent, does not:

```
$ export CMDLINER_LEGACY_PREFIXES=1
$ stanc -<TAB>
Usage: %%NAME%% [--help] [OPTION]… [MODEL_FILE]
stanc: option - ambiguous and could be one of --O0, --O1, --Oexperimental,
```

In some ways this seems silly -- it's listing all the valid completions in the error message!

To directly see the issue, the same error arises if you 

`$ CMDLINER_LEGACY_PREFIXES=1 stanc --__complete --__complete=-`

(it works fine if there are prefixes used before the --_complete= item, or if it's unambiguous, like in stanc this does offer valid completions even though the argument names are `auto-format` and `canonicalize`: `$ CMDLINER_LEGACY_PREFIXES=1 stanc --__complete --__complete=--canon=<TAB>`)

---

In particular, this means that the patch you suggest [here](https://github.com/ocaml/opam/issues/6849#issuecomment-3802583384) and [here](https://github.com/ocaml/dune/issues/12515#issuecomment-3802602856) breaks completion for more or less all options, as `-` will always be highly ambiguous. Some sort of special-casing seems worthwhile?

WardBrian:
Closes #262

### DIFF
--- a/src/cmdliner_cline.ml
+++ b/src/cmdliner_cline.ml
@@ -202,7 +202,7 @@ let parse_opt_args
           in
           let cline = Cmdliner_def.Cline.add arg_info arg cline in
           loop errs (k + 1) comp cline pargs args
-      | Error `Not_found when for_completion ->
+      | Error (`Not_found | `Ambiguous) when for_completion ->
           if not is_completion then
             (* Drop the data, if the user thought this was an opt with
                an argument this may confuse positional args but there's


WardBrian:
@dbuenzli I never circled back on this but this did seem sufficient for #262, no other issues arose while using it in my opam branches. Mind merging?

dbuenzli:
Yes, the fix looks correct we obviously need to complete when the trie lookup is ambiguous. Thanks, your patch is in.

PR #267, opened 2026-04-19
Use dedicated Groff escape sequence for hyphens

### LINKED ISSUE #254: Groff output breaks Lexgrog
I’m packaging an app that uses Cmdliner for Debian, and generating the manpages by piping `<command> --help=groff` to my manpage files.
It works great, the manpages can properly be read, but the Debian package linter, `lintian`, uses `lexgrog` to parse the manpages for some metadata. However, `lexgrog` doesn’t like escape characters in the `NAME` section. Cmdliner is generating `\N'45'` for the regular dash `-` character, which breaks this tool (and thus quite a few Debian packaging tools). Would it be possible to generate `\-` instead? (or `-` within the text and `\-` for command-line options, but that one may be too tricky)

For now, I’m piping the output to `sed "s/\\N'45'/\\-/g"`, but was hoping for a more “legacy” handling of dashes that works well with groff parsers. What do you think?

bensmrs:
This makes man-db and Debian tools happy, as they may have pretty strict parsing rules when it comes to hyphens.

Closes: #254

### DIFF
--- a/src/cmdliner_manpage.ml
+++ b/src/cmdliner_manpage.ml
@@ -374,7 +374,10 @@ let markup_to_groff ~errs b s =
   | false -> Buffer.add_substring b s start (stop - start + 1)
   in
   let need_escape = function '.' | '\'' | '-' | '\\' -> true | _ -> false in
-  let escape b c = Printf.bprintf b "\\N'%d'" (Char.code c) in
+  let escape b = function
+  | '-' -> Printf.bprintf b "\\-"
+  | c -> Printf.bprintf b "\\N'%d'" (Char.code c)
+  in
   let rec end_text start i = Buffer.add_string b "\\fR"; loop start i
   and loop start i =
     if i > max_i then flush start max_i else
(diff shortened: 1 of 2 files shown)


src/cmdliner_manpage.ml
```diff
@@ -374,7 +374,10 @@ let markup_to_groff ~errs b s =
   | false -> Buffer.add_substring b s start (stop - start + 1)
   in
   let need_escape = function '.' | '\'' | '-' | '\\' -> true | _ -> false in
-  let escape b c = Printf.bprintf b "\\N'%d'" (Char.code c) in
+  let escape b = function
+  | '-' -> Printf.bprintf b "\\-"
```
dbuenzli, line 378: Could just please add a comment so that we don't accidentaly remove the line in the future. Something like `(* use this escape form to please lexgrog *)`

dbuenzli (approved):

dbuenzli:
Thanks, your patch is in.

bensmrs:
Oh I missed your comment, sorry about that and thanks

dbuenzli:
No problem, there was suddendly an urge to release :–) So I amended the patch.

